Issue #101, opened 2015-01-20
Use tls min version flags similar to nsqd / set tls max version


twmb:
To mirror https://github.com/bitly/nsq/pull/518, which changes -tls-min-version from the format "tls10" to "tls.10", and which also set tls max version to get TLS_FALLBACK_SCSV

mreiferson:
:+1:

PR #102, opened 2015-01-21
tls updates: match nsqd min version flag, set default max/min (per #101)

twmb:
- Matches nsqd's -tls-min-version flag format, which was changed in bitly/nsqd#518
- Defaults tls min version to TLS1.0
- Sets tls max version to enable TLS_FALLBACK_SCSV prior to Go 1.5
    - see https://go-review.googlesource.com/#/c/1776/ (thanks @jehiah)


twmb:
RFR @mreiferson

PR #104, opened 2015-01-27
consumer: Only add host address once on connection.

ryanslade:
When specifying `nsqd` addresses directly, upon reconnecting, the address was being duplicated in `nsqdTCPAddrs`


mreiferson:
thanks!

PR #106, opened 2015-02-02
consumer: Reset backoff counter when no connections left.

ryanslade:
This fixes the following scenario:

1. Consumer is connected to one NSQD host using ConnectToNSQD.
2. Consumer receives message and handler returns an error.
3. Consumer backs off.
4. DisconnectFromNSQD called before backoff closure wakes up.
5. Backoff closure wakes up, sees no connections, returns.
6. Consumer connects to NSQD host using ConnectToNSQD.
7. backoffCounter still > 0 so never sends RDY count (in
maybeUpdateRDY).
8. Consumer is now stuck and never sends RDY > 0.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -660,32 +660,7 @@ func (r *Consumer) startStopContinueBackoff(conn *Conn, success bool) {
 		// start or continue backoff
 		backoffDuration := r.backoffDurationForCount(r.backoffCounter)
 		atomic.StoreInt64(&r.backoffDuration, backoffDuration.Nanoseconds())
-		time.AfterFunc(backoffDuration, func() {
-			var choice *Conn
-
-			atomic.StoreInt64(&r.backoffDuration, 0)
-
-			// pick a random connection to test the waters
-			var i int
-			conns := r.conns()
-			if len(conns) == 0 {
-				return
-			}
-			idx := r.rng.Intn(len(conns))
-			for _, c := range conns {
-				if i == idx {
-					choice = c
-					break
-				}
-				i++
-			}
-
-			r.log(LogLevelWarning,
-				"(%s) backoff timeout expired, sending RDY 1",
-				choice.String())
-			// while in backoff only ever let 1 message at a time through
-			r.updateRDY(choice, 1)
-		})
+		time.AfterFunc(backoffDuration, r.backoff)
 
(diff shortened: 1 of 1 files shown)


consumer.go
```diff
@@ -669,6 +669,11 @@ func (r *Consumer) startStopContinueBackoff(conn *Conn, success bool) {
 			var i int
 			conns := r.conns()
 			if len(conns) == 0 {
+				r.backoffMtx.Lock()
+				if r.backoffCounter > 0 {
+					r.backoffCounter--
+				}
+				r.backoffMtx.Unlock()
```
mreiferson: Thanks for identifying this bug.

Unfortunately, I'm not convinced this is the right fix.  I *think* the correct behavior should be for this to loop until it has a connection (also paying attention to an exit signal), and then send `RDY 1`.

Thoughts @jehiah

jehiah: If we treat backoff as backing off from downstream systems, then i agree. It should persist beyond upstream nsqd reconnects

Since this code block is in a `time.After` we should be able to refactor to just re-set that function recursively (and have it end on exit flags appropriately).

ryanslade: I should have some time to work on a fix tomorrow (GMT).

consumer.go
```diff
@@ -697,6 +672,34 @@ func (r *Consumer) startStopContinueBackoff(conn *Conn, success bool) {
...
+
+	// pick a random connection to test the waters
+	conns := r.conns()
+	if len(conns) == 0 {
+		// backoff again
+		r.backoffMtx.RLock()
+		backoffDuration := r.backoffDurationForCount(r.backoffCounter)
+		r.backoffMtx.RUnlock()
+		atomic.StoreInt64(&r.backoffDuration, backoffDuration.Nanoseconds())
+		time.AfterFunc(backoffDuration, r.backoff)
```
mreiferson: I don't think we need to wait for the entire (previously expired) backoff interval - I think we can just use some (small) constant sleep like `1s`.

Simplifies this inner block of code, too.

mreiferson:
this LGTM, mind squashing the commits down?

ryanslade:
Squashed :)

mreiferson:
thanks!

PR #107, opened 2015-02-04
Fix bounds err when removing lookupds

andreas:
When removing lookupds, `r.lookupdQueryIndex` may become greater than `len(r.lookupdHTTPAddrs)`. This can cause a panic by index out of bounds error.


mreiferson:
:+1:

PR #109, opened 2015-02-05
Producer: Adds a Ping method

dcarney:
Adding a Ping method to the Producer allows for testing a Producer's nsqd connection w/o publishing any messages to a topic.

We've found it to be useful for failing fast when configuring Producers (since otherwise, the connections are made lazily on Publish), so I thought I'd see if it's a PR you want to merge in.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -82,6 +82,11 @@ func NewProducer(addr string, config *Config) (*Producer, error) {
 	return p, nil
 }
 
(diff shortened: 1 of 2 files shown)


dcarney:
In the midst of getting all my commits pushed to this PR, I realized a better way to accomplish this.  As such, I'm going to close/abandon this PR.

PR #113, opened 2015-02-09
consumer: jitter might have no effect

judwhite:
please review if this is correct behavior

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -73,9 +73,6 @@ type Conn struct {
 	r io.Reader
 	w io.Writer
 
-	backoffCounter int32
-	rdyRetryTimer  *time.Timer
-
 	cmdChan         chan *Command
 	msgResponseChan chan *msgResponse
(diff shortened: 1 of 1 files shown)


mreiferson:
Can you explain these changes? Not sure this is desirable...

judwhite:
ticker starts when it's created. when it's created before waiting on jitter the jitter has no effect, unless jitter is greater than the tick interval.

example:

	start := time.Now()

	ticker := time.NewTicker(1500 * time.Millisecond)
	time.Sleep(1000 * time.Millisecond)

	<-ticker.C

	fmt.Println(time.Now().Sub(start))

mreiferson:
ahh, that makes more sense :+1:, really nice catch!

mreiferson:
mind squashing down?  thanks for this...

judwhite:
reopened as https://github.com/bitly/go-nsq/pull/114

PR #115, opened 2015-02-10
remove unnecessary check on pendingOk

judwhite:
see above if statement: if ok || pendingOk { r.mtx.Unlock() return ErrAlreadyConnected }


mreiferson:
yep yep, thanks!

PR #119, opened 2015-02-18
producer: Adds a Ping method in order to validate an nsqd connection

dcarney:
Adds a Ping method to the Producer, which allows for testing a Producer's nsqd connection w/o publishing any messages to a topic. 

I've found it to be useful for failing fast when configuring Producers that might not Publish immediately (since otherwise, the connections are only made lazily on Publish).

Not thrilled with the method name "Ping", so am totally open to other suggestions.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -91,6 +91,23 @@ func NewProducer(addr string, config *Config) (*Producer, error) {
 	return p, nil
 }
 
+// Ping causes the Producer to connect to it's configured nsqd (if not already
+// connected) and send a `Nop` command, returning any error that might occur.
+//
+// This method can be used to verify that a newly-created Producer instance is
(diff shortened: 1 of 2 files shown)


mreiferson:
My only comment is - let's actually make this send a `NOP` command to the `nsqd`?

mreiferson:
works for me - do you mind squashing down to one commit?

@jehiah any thoughts before we merge?

jehiah:
:thumbsup:

dcarney:
@mreiferson Squashed it.  Thanks!

mreiferson:
hmm, looks like some other recently merged code is conflicting - needs a rebase now, sorry!

dcarney:
No problem.  Rebased it, and also removed an unecessary "fmt" import.

PR #120, opened 2015-02-21
message: atomic races

mreiferson:
there was the possibility of a race since we were separately checking/storing

cc @schmichael

### DIFF
--- a/message.go
+++ b/message.go
@@ -65,11 +65,10 @@ func (m *Message) HasResponded() bool {
 // Finish sends a FIN command to the nsqd which
 // sent this message
 func (m *Message) Finish() {
-	if m.HasResponded() {
+	if !atomic.CompareAndSwapInt32(&m.responded, 0, 1) {
 		return
 	}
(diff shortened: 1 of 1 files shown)


mreiferson:
RFR @jehiah

message.go
```diff
@@ -65,17 +65,16 @@ func (m *Message) HasResponded() bool {
...
 	}
 	m.Delegate.OnFinish(m)
-	atomic.StoreInt32(&m.responded, 1)
 }
 
 // Touch sends a TOUCH command to the nsqd which
 // sent this message
 func (m *Message) Touch() {
-	if m.HasResponded() {
+	if !atomic.CompareAndSwapInt32(&m.responded, 0, 1) {
```
jehiah: just `LoadInt32` here since we don't want `Touch()` to mark as responded.

mreiferson: ahh good catch

mreiferson:
fixed

PR #123, opened 2015-03-05
add jittered backoff

hden:
Based-on http://www.awsarchitectureblog.com/2015/03/backoff.html

Any thoughts?

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
 	"os"
 	"reflect"
 	"strconv"
@@ -24,6 +25,24 @@ type configHandler interface {
 
 type defaultsHandler interface {
 	SetDefaults(c *Config) error
+}
+
+// Define backoffStrategy
+type BackoffStrategy interface {
+   Calculate(attempt int) time.Duration
+}
+
+type ExponentialStrategy struct {
+	config *Config
+}
+
+func (s ExponentialStrategy) Calculate(attempt int) time.Duration {
+	backoffDuration := s.config.BackoffMultiplier *
+		time.Duration(math.Pow(2, float64(attempt)))
+	if backoffDuration > s.config.MaxBackoffDuration {
+		backoffDuration = s.config.MaxBackoffDuration
+	}
+	return backoffDuration
 }
 
 // Config is a struct of NSQ options
@@ -103,6 +122,8 @@ type Config struct {
 
 	// Maximum amount of time to backoff when processing fails 0 == no backoff
 	MaxBackoffDuration time.Duration `opt:"max_backoff_duration" min:"0" max:"60m" default:"2m"`
+	// Backoff strategy, defaults to exponential backoff. Overwrite this to define alternative backoff algrithms.
+	BackoffStrategy BackoffStrategy
 
 	// The server-side message timeout for messages delivered to this client
 	MsgTimeout time.Duration `opt:"msg_timeout" min:"0"`
@@ -118,6 +139,7 @@ func NewConfig() *Config {
 	c := &Config{}
 	c.configHandlers = append(c.configHandlers, &structTagsConfig{}, &tlsConfig{})
 	c.initialized = true
+	c.BackoffStrategy = BackoffStrategy(ExponentialStrategy{c})
 	if err := c.setDefaults(); err != nil {
 		panic(err.Error())
 	}
--- a/config.go
+++ b/config.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"math"
+	"math/rand"
 	"os"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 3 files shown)


mreiferson:
this is very cool, thanks for contributing!

config.go
```diff
@@ -52,8 +52,6 @@ type Config struct {
 	// Maximum duration when REQueueing (for doubling of deferred requeue)
 	MaxRequeueDelay     time.Duration `opt:"max_requeue_delay" min:"0" max:"60m" default:"15m"`
 	DefaultRequeueDelay time.Duration `opt:"default_requeue_delay" min:"0" max:"60m" default:"90s"`
-	// Unit of time for calculating consumer backoff
-	BackoffMultiplier time.Duration `opt:"backoff_multiplier" min:"0" max:"60m" default:"1s"`
```
mreiferson: Unfortunately we can't just remove this (backwards compatibility) - so it might make sense to just use `BackoffMultipler` as `MinBackoffDuration`?

consumer.go
```diff
@@ -815,8 +815,14 @@ func (r *Consumer) onConnClose(c *Conn) {
...
-	backoffDuration := r.config.BackoffMultiplier *
-		time.Duration(math.Pow(2, float64(count)))
+	// Jittered Backoff
+	// http://www.awsarchitectureblog.com/2015/03/backoff.html
+	if 0 == r.backoffDuration {
+		return r.config.MinBackoffDuration
+	}
+	base := int(r.config.MinBackoffDuration)
+	sleep := int(r.backoffDuration)
+	backoffDuration := time.Duration((r.rng.Intn(sleep * 3) - base) + base)
```
mreiferson: this implements the "decorrelated" algorithm from the article, since we already have this setup to be based on "attempt" (`count`) it might be best to use the "full jitter" algorithm, which uses attempts.

They seem to perform basically the same.

Alternatively, we should probably rename this function and remove its parameter since it isn't being used.

consumer.go
```diff
@@ -815,8 +815,14 @@ func (r *Consumer) onConnClose(c *Conn) {
 }
 
 func (r *Consumer) backoffDurationForCount(count int32) time.Duration {
-	backoffDuration := r.config.BackoffMultiplier *
-		time.Duration(math.Pow(2, float64(count)))
+	// Jittered Backoff
+	// http://www.awsarchitectureblog.com/2015/03/backoff.html
```
mreiferson: would be great to comment on the *specifc* algorithm we're implementing

consumer.go
```diff
@@ -815,8 +815,14 @@ func (r *Consumer) onConnClose(c *Conn) {
...
+	if 0 == r.backoffDuration {
```
mreiferson: would prefer this logic reversed, i.e. `r.backoffDuration == 0`

hden:
@mreiferson done

mreiferson:
I guess my only other question is - should we make these backoff modes optional and configurable (i.e. expose an interface that the end-user can implement themselves plus provide default implementations for "exponential" and "full jitter")

@jehiah what do you think?

hden:
Maybe we can pick a reasonable default and [expose the possibilities](https://extensiblewebmanifesto.org) to overwrite the algorithm in the userland?

mreiferson:
Exactly, I'm thinking something like"

```go
type BackoffStrategy interface {
   Calculate(attempt int) time.Duration
}

type FullJitterStrategy struct {}
func (s FullJitterStrategy) Calculate(attempt int) time.Duration { ... }

type ExponentialStrategy struct {}
func (s ExponentialStrategy) Calculate(attempt int) time.Duration { ... }
```


And we would add a property to `Config` and default it to `ExponentialStrategy`, to maintain backwards compatibility.

What do you think?

hden:
Awesome. :+1:  Should I add something like a `backoffConfig` at [L119](https://github.com/bitly/go-nsq/blob/master/config.go#L119) or just initiate a `ExponentialStrategy` in the composite literal?

mreiferson:
Yea, add a new exported property somewhere around https://github.com/bitly/go-nsq/blob/master/config.go#L105?

config.go
```diff
@@ -26,6 +27,24 @@ type defaultsHandler interface {
 	SetDefaults(c *Config) error
 }
 
+// Define backoffStrategy
+type BackoffStrategy interface {
+   Calculate(attempt int) time.Duration
+}
+
+type ExponentialStrategy struct {
+	config *Config
```
mreiferson: if `config` isn't exported, you won't actually be able to instantiate it outside this package.

config.go
```diff
@@ -26,6 +27,24 @@ type defaultsHandler interface {
 	SetDefaults(c *Config) error
 }
 
+// Define backoffStrategy
+type BackoffStrategy interface {
+   Calculate(attempt int) time.Duration
+}
+
+type ExponentialStrategy struct {
```
mreiferson: I think we should also provide `FullJitterStrategy` as an exported part of the package?

config.go
```diff
@@ -118,6 +139,7 @@ func NewConfig() *Config {
 	c := &Config{}
 	c.configHandlers = append(c.configHandlers, &structTagsConfig{}, &tlsConfig{})
 	c.initialized = true
+	c.BackoffStrategy = BackoffStrategy(ExponentialStrategy{c})
```
mreiferson: I don't think you need the `BackoffStrategy(...)` cast

mreiferson:
This is looking really good, almost there, thanks!

config.go
```diff
@@ -26,6 +28,41 @@ type defaultsHandler interface {
 	SetDefaults(c *Config) error
 }
 
+// Define backoffStrategy
+type BackoffStrategy interface {
+   Calculate(attempt int) time.Duration
```
mreiferson: looks like this needs `go fmt`?

config.go
```diff
@@ -26,6 +28,41 @@ type defaultsHandler interface {
...
+		backoffDuration = s.Config.MaxBackoffDuration
+	}
+	return backoffDuration
+}
+
+// Full Jittered Backoff
+// http://www.awsarchitectureblog.com/2015/03/backoff.html
+type FullJitterStrategy struct {
+	Config *Config
+	Rng *rand.Rand
```
mreiferson: my last question is, can we think of any way to avoid needing to instantiate these strategies with `Config` (and in this case, `Rng`)?

For `Rng`, we could use a private `sync.Once` to create it the first time `Calculate` is called.

Any ideas for `Config`?  I want it to be easy and obvious how to instantiate these properly *without* resorting to `NewXStrategy()` funcs.

mreiferson:
closing in favor of #125

PR #124, opened 2015-03-19
Bind outbound connection to address from configuration

yashkin:
For security reasons it would be great to support such feature as
binding outbound address, so nsqd could be firewalled on accepting
certain addresses, that reasonable in case when nsq-consumers or
producers are placed on machines with multiple interfaces.

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net"
 	"os"
 	"reflect"
 	"strconv"
@@ -39,9 +40,15 @@ type Config struct {
 	// used to Initialize, Validate
 	configHandlers []configHandler
 
+	DialTimeout time.Duration `opt:"dial_timeout" default:"1s"`
+
 	// Deadlines for network reads and writes
 	ReadTimeout  time.Duration `opt:"read_timeout" min:"100ms" max:"5m" default:"60s"`
 	WriteTimeout time.Duration `opt:"write_timeout" min:"100ms" max:"5m" default:"1s"`
+
+	// LocalAddr is the local address to use when dialing an nsqd.
+	// If empty, a local address is automatically chosen.
+	LocalAddr net.Addr `opt:"local_addr"`
 
 	// Duration between polling lookupd for new producers, and fractional jitter to add to
 	// the lookupd pool loop. this helps evenly distribute requests even if multiple consumers
@@ -456,6 +463,8 @@ func coerce(v interface{}, typ reflect.Type) (reflect.Value, error) {
 		v, err = coerceBool(v)
 	case "time.Duration":
 		v, err = coerceDuration(v)
+	case "net.Addr":
+		v, err = coerceAddr(v)
 	default:
 		v = nil
 		err = errors.New(fmt.Sprintf("invalid type %s", typ.String()))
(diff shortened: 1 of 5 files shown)


mreiferson:
I'm not necessarily opposed to this change but can't this be addressed outside of `go-nsq` by configuring routing tables?

twmb:
This will try to bind all outgoing connections to the same address. Unless I'm mistaken, if you set LocalAddr, this will fail when you have more than one server you're trying to connect to, as this will try to open two connections with the same source addr.

yashkin:
mreiferson, imagine machine with two NICs, when you start nsq listener or consumer, ip address of which NIC they will use? The answer is unpredictable, probably, it's possible to solve issue with iptables and masquerading of **all** outgoing ip addresses, but it's tricky, isn't it?

twmb, if you look into `consumer.go` you will find out, that it uses map of connections, where key is address of nsq daemon, so you can easily connect to multiple servers. Either, if you will use `LocalAddr` with port equal to *0*, like `config.Set("local_addr", "127.0.0.1:0")` os find a free one and bind your client to that address, so there is no problems with multiple connections (they all will use different endpoints).

consumer_test.go
```diff
@@ -187,6 +192,11 @@ func consumerTest(t *testing.T, cb func(c *Config)) {
 		t.Fatal("should not be able to connect to the same NSQ twice")
 	}
 
+	conn := q.conns()[0]
+	if conn.conn.LocalAddr().String() != laddr {
+		t.Fatal("connection should be binded to the specified address:", conn.conn.LocalAddr())
```
twmb: should be `bound`

twmb:
Cool, I hadn't read that yet. This looks fine to me other than `binded` should be `bound` in error messages. I'd change `ConnTimeout` to `DialTimeout` as the timeout is only used in dialing.

yashkin:
Thanks, fixed.

mreiferson:
> mreiferson, imagine machine with two NICs, when you start nsq listener or consumer, ip address of which NIC they will use? The answer is unpredictable, probably, it's possible to solve issue with iptables and masquerading of all outgoing ip addresses, but it's tricky, isn't it?

Still feels wrong to me.  I don't disagree that this patch addresses the issue, my concern is that this approach implies that _all_ of the client libraries need the same patch.  Whereas, if you handled this at the networking layer, it would apply to _anything_ connecting to nsqd.

Curious to hear @jehiah's thoughts...

yashkin:
If you leave `LocalAddr` parameter equals to `nil` clients would behave the same as before. It's an option, not a requirement.

jehiah:
certainly won't be needed in most cases but i'm fine exposing this. :thumbsup:

mreiferson:
alright alright

config.go
```diff
@@ -476,14 +485,16 @@ func valueTypeCoerce(v interface{}, typ reflect.Type) reflect.Value {
...
 		tval.SetFloat(val.Float())
+	default:
+		tval.Set(val)
 	}
 	return tval
 }
 
 func coerceString(v interface{}) (string, error) {
-	switch v.(type) {
+	switch v := v.(type) {
```
mreiferson: let's not shadow `v` here (and all the other instances below)

conn.go
```diff
@@ -139,8 +139,13 @@ func (c *Conn) getLogger() (logger, LogLevel, string) {
 
 // Connect dials and bootstraps the nsqd connection
 // (including IDENTIFY) and returns the IdentifyResponse
-func (c *Conn) Connect() (*IdentifyResponse, error) {
-	conn, err := net.DialTimeout("tcp", c.addr, time.Second)
+func (c *Conn) Connect() (_ *IdentifyResponse, err error) {
```
mreiferson: prefer not to have named returns

consumer_test.go
```diff
@@ -223,3 +233,9 @@ func consumerTest(t *testing.T, cb func(c *Config)) {
 		t.Fatal("failed message not done")
 	}
 }
+
+func port() string {
```
mreiferson: should we listen on `:0`, read the port, close, then return that assumed open port?

config.go
yashkin: it's golang idiom -reuse name of variables in a type switch statements
https://golang.org/doc/effective_go.html#type_switch

mreiferson: TIL - thanks!

mreiferson:
@yashkin thanks, would you mind squashing down to one commit and we'll merge?

PR #125, opened 2015-03-22
consumer: backoff strategies

mreiferson:
refactoring of the work in #123

RFR @jehiah

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,11 +7,14 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math"
+	"math/rand"
 	"net"
 	"os"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


jehiah:
code looks good (Aside from integration test failures)

mreiferson:
updated

jehiah:
LGTM

mreiferson:
:hammer:

twmb:
It nice be beneficial to document that the code calls `setConfig` so that implementors can also use the function, the same way `flag` does `IsBoolFlag`.

mreiferson:
@twmb I made those explicitly private, I don't want others relying on that method/interface/etc.  My expectation is that user-implemented `BackoffStrategy` would use their own parameters/config.

PR #126, opened 2015-03-23
Allow BackoffStrategy to be set via flag, document BackoffStrategy.SetConfig

twmb:
- Documents that go-nsq calls SetConfig on BackoffStrategy, allowing
  implementors to use config values in their own Calculate implementation.
- Allows full_jitter and exponential strategies to be set via the
  backoff_strategy flag.

### DIFF
--- a/config.go
+++ b/config.go
@@ -119,7 +119,7 @@ type Config struct {
 	DefaultRequeueDelay time.Duration `opt:"default_requeue_delay" min:"0" max:"60m" default:"90s"`
 
 	// Backoff strategy, defaults to exponential backoff. Overwrite this to define alternative backoff algrithms.
-	BackoffStrategy BackoffStrategy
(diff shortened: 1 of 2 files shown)


mreiferson:
at a quick glance this looks great - curious to hear your thoughts on [my response](https://github.com/bitly/go-nsq/pull/125#issuecomment-85227966) to your question in #125

twmb:
I just saw it. I don't mind much either way. If a person wanted to use the config they could just
```
cfg := nsq.NewConfig()
myStrat.cfg = cfg
cfg.Set("backoff_strategy", myStrat)
```
Because it's so easy to use the config regardless, it doesn't make much sense to expose `setConfig` and document it, especially because it also clutters the api.

mreiferson:
Agreed - let's remove those changes then, thanks!

mreiferson:
LGTM

PR #127, opened 2015-03-23
when in backoff, make sure that the chosen connection isn't closing

jnewmano:
Backoff chooses a random connection and tries to update the ready status; however, if the connection is closing, updateRdy will return immediately, leaving the backoffCounter in a bad state. The side effect is that maybeUpdateRdy will never update the ready count due to the incorrect backoffCounter and the consumer will never again consume messages.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -691,9 +691,9 @@ func (r *Consumer) startStopContinueBackoff(conn *Conn, success bool) {
 }
 
 func (r *Consumer) backoff() {
-	atomic.StoreInt64(&r.backoffDuration, 0)
 
 	if atomic.LoadInt32(&r.stopFlag) == 1 {
+		atomic.StoreInt64(&r.backoffDuration, 0)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


jnewmano:
It seems like a regression on #106

jnewmano:
It may be better (and avoid a potential race) to have updateRdy return an error if c.IsClosing() and have backoff handle the error appropriately.

mreiferson:
Was going to suggest the same, want to give that a try?

Thanks for finding this!

jnewmano:
I made the switch to checking the returned error. 

On consumer.go line 694 clears backoffDuration perhaps momentarily.  It may be better to move it to the end of backoff(), after backoff is successful? This would avoid mistakenly sending false to a call to inBackoffBlock()

https://github.com/jnewmano/go-nsq/blob/master/consumer.go#L694

consumer.go
```diff
@@ -709,11 +709,29 @@ func (r *Consumer) backoff() {
 	idx := r.rng.Intn(len(conns))
 	choice := conns[idx]
 
+	// make sure the chosen connection isn't closing
+	if choice.IsClosing() {
```
mreiferson: I don't think we need this block anymore, do we?

consumer.go
```diff
@@ -709,11 +709,29 @@ func (r *Consumer) backoff() {
...
+		return
+	}
+
 	r.log(LogLevelWarning,
 		"(%s) backoff timeout expired, sending RDY 1",
 		choice.String())
 	// while in backoff only ever let 1 message at a time through
-	r.updateRDY(choice, 1)
+	err := r.updateRDY(choice, 1)
+
```
mreiferson: drop this whitespace :cop:

consumer.go
```diff
@@ -709,11 +709,29 @@ func (r *Consumer) backoff() {
...
+	if err != nil {
```
mreiferson: let's add a log line for this case...

mreiferson:
>On consumer.go line 694 clears backoffDuration perhaps momentarily. It may be better to move it to the end of backoff(), after backoff is successful? This would avoid mistakenly sending false to a call to inBackoffBlock()

SGTM

mreiferson:
LGTM

PR #129, opened 2015-03-25
Allow backoff to be completely disabled in consumer

jcoene:
This pull request adds a `backoff_enabled` option to Config with the default value being set to true. When disabled, the consumer will never back off.

This is different from setting a `backoff_duration` of 0, as in that case the consumer will still send `RDY 0` to nsqd, decreasing throughput significantly. Disabling the `backoff_enabled` option allows us to sustain high throughput when messages are failing.

There were quite a few ways to implement this - I went with the simplest to read and probably most efficient rather than complicating the existing backoff logic and/or inflating the BackoffStrategy interface. The new configuration option has a nice side effect of not changing the existing behavior for anyone who might have `backoff_duration` set to 0 then suddenly see their throughput increase under error conditions.

I'm not sure how to build an automated test for this.

### DIFF
--- a/config.go
+++ b/config.go
@@ -118,6 +118,8 @@ type Config struct {
 	MaxRequeueDelay     time.Duration `opt:"max_requeue_delay" min:"0" max:"60m" default:"15m"`
 	DefaultRequeueDelay time.Duration `opt:"default_requeue_delay" min:"0" max:"60m" default:"90s"`
 
+	// Backoff enabled, defaults to true. Setting this to false disables backoff entirely.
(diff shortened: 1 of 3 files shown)


jehiah:
@jcoene this can already be controlled already by using [`msg.RequeueWithoutBackoff()`](http://godoc.org/github.com/bitly/go-nsq#Message.RequeueWithoutBackoff) on message errors instead of `msg.Requeue()` or returning an error from the handler.

Can you share about how that does or doesn't work for your situation?

jcoene:
Ah, that would work just fine. Thanks for pointing me in the right direction!

PR #133, opened 2015-04-08
producer: fix connection race

jeddenlea:
Prior to this change, it is possible for more than one goroutine to
check:

	if atomic.LoadInt32(&w.state) != StateConnected {

...thereafter calling w.connect.  w.connect first grabs a lock on
w.guard, and then asserts that it be the only goroutine attempting a
connection:

	if !atomic.CompareAndSwapInt32(&w.state, StateInit, StateConnected) {
		return ErrNotConnected
	}

This means if there are two parallel actions which attempt to initiate
the connection, one of them will return ErrNotConnected if the first
succesfully connected.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -243,7 +243,11 @@ func (w *Producer) connect() error {
 		return ErrStopped
 	}
 
-	if !atomic.CompareAndSwapInt32(&w.state, StateInit, StateConnected) {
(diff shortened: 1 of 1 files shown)


mreiferson:
LGTM, nice catch!

Thanks!

Issue #136, opened 2015-04-29
Is there any reasons to use go-simplejson rather than built-in json package?


chzyer:
Hi, guys.
While I'm reading the source in `api_request.go`, I found that it use "simplejson" package to unmarshal json in the method called `apiRequestNegotiateV1`.

Why not just define a struct and pass a pointer which address to it into `json.Unmarshal`.
I think that may improve the readability of the code and may not affect the performance too much.

mreiferson:
@chzyer you're right, there's no compelling reason to depend on that package.  I would merge a patch that makes those changes.

mreiferson:
see #137

PR #137, opened 2015-04-29
remove go-simplejson dependency

chzyer:
Solution for #136.
Remove "simple-json" dependency completely.
Please let me know if anything I can do.

### DIFF
--- a/api_request.go
+++ b/api_request.go
@@ -1,14 +1,13 @@
 package nsq
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
 	"time"
-
-	"github.com/bitly/go-simplejson"
 )
 
 type deadlinedConn struct {
@@ -39,40 +38,53 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
 	return transport
 }
 
-func apiRequestNegotiateV1(method string, endpoint string, body io.Reader) (*simplejson.Json, error) {
+type wrappedResp struct {
+	Status     string      `json:"status_txt"`
+	StatusCode int         `json:"status_code"`
+	Data       interface{} `json:"data"`
+}
+
+// stores the result in the value pointed to by ret(must be a pointer)
+func apiRequestNegotiateV1(method string, endpoint string, body io.Reader, ret interface{}) error {
 	httpclient := &http.Client{Transport: newDeadlineTransport(2 * time.Second)}
 	req, err := http.NewRequest(method, endpoint, body)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	req.Header.Add("Accept", "application/vnd.nsq; version=1.0")
 
 	resp, err := httpclient.Do(req)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	respBody, err := ioutil.ReadAll(resp.Body)
 	resp.Body.Close()
 	if err != nil {
-		return nil, err
+		return err
 	}
+
 	if resp.StatusCode != 200 {
-		return nil, fmt.Errorf("got response %s %q", resp.Status, respBody)
+		return fmt.Errorf("got response %s %q", resp.Status, respBody)
 	}
 
 	if len(respBody) == 0 {
 		respBody = []byte("{}")
 	}
 
-	data, err := simplejson.NewJson(respBody)
-	if err != nil {
-		return nil, err
+	if resp.Header.Get("X-NSQ-Content-Type") == "nsq; version=1.0" {
+		return json.Unmarshal(respBody, ret)
 	}
 
-	if resp.Header.Get("X-NSQ-Content-Type") == "nsq; version=1.0" {
-		return data, nil
+	wResp := &wrappedResp{
+		Data: ret,
 	}
-	return data.Get("data"), nil
+
+	if err = json.Unmarshal(respBody, wResp); err != nil {
+		return err
+	}
+
+	// wResp.StatusCode here is equal to resp.StatusCode, so ignore it
+	return nil
 }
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -431,6 +431,21 @@ func (r *Consumer) nextLookupdEndpoint() string {
 	return u.String()
 }
 
+type LookupResp struct {
```
mreiferson: let's make this private

consumer.go
```diff
@@ -431,6 +431,21 @@ func (r *Consumer) nextLookupdEndpoint() string {
...
+	Channels  []string    `json:"channels"`
+	Producers []*PeerInfo `json:"producers"`
+	Timestamp int64       `json:"timestamp"`
+}
+
+type PeerInfo struct {
```
mreiferson: this too (private)

consumer.go
```diff
@@ -440,28 +455,17 @@ func (r *Consumer) queryLookupd() {
 
 	r.log(LogLevelInfo, "querying nsqlookupd %s", endpoint)
 
-	data, err := apiRequestNegotiateV1("GET", endpoint, nil)
+	data := new(LookupResp)
```
mreiferson: we avoid `new(...)`, let's use `var data LookupResp` and then `&data` below, please

mreiferson:
nice work! thanks for cleaning this up!

api_request.go
```diff
@@ -39,40 +38,52 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
...
 }
 
-func apiRequestNegotiateV1(method string, endpoint string, body io.Reader) (*simplejson.Json, error) {
+type generalResp struct {
+	Status     string      `json:"status_txt"`
+	StatusCode int         `json:"statuc_code"`
+	Data       interface{} `json:"data"`
+}
+
+func apiRequestNegotiateV1(method string, endpoint string, body io.Reader, ret interface{}) error {
```
mreiferson: this never uses `ret` - should we just keep the signature similar and actually `return` the data rather than pass it in?

chzyer: what did you means "this never uses ret"?

mreiferson: ahh, sorry, in the non `nsq; version=1.0` case, the return value is lost because we need to use `generalResp`

api_request.go
```diff
@@ -39,40 +38,52 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
 	return transport
 }
 
-func apiRequestNegotiateV1(method string, endpoint string, body io.Reader) (*simplejson.Json, error) {
+type generalResp struct {
```
mreiferson: a better name for this would be `wrappedResp`

api_request.go
chzyer: What did you mean "this never uses ret"?
`apiRequestNegotiateV1` may not know the type of response, so it couldn't return the data directly.

mreiferson: see above comment?

chzyer: Did you means that `ret` will be not change in the non nsq; version=1.0 case?

api_request.go
```diff
@@ -39,40 +38,52 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
...
-	if err != nil {
-		return nil, err
+	if resp.Header.Get("X-NSQ-Content-Type") == "nsq; version=1.0" {
+		return json.Unmarshal(respBody, ret)
 	}
 
-	if resp.Header.Get("X-NSQ-Content-Type") == "nsq; version=1.0" {
-		return data, nil
+	gResp := &generalResp{
+		Data: ret,
```
mreiferson: ahhh, sorry, I missed this!

mreiferson: does this actually work?

chzyer: Yes :)
A `interface{}` value contains a type value and a pointer value.
So It actually set the value to the memory where `ret` address.
http://play.golang.org/p/Xqu5YY8e2L

api_request.go
mreiferson: fascinating, thanks!

mreiferson:
If this test run passes, we can squash these commits down to a single one and we're ready to go :100:

chzyer:
ok. I force pushed a new commit :)

api_request.go
```diff
@@ -39,40 +38,52 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
 	return transport
 }
 
-func apiRequestNegotiateV1(method string, endpoint string, body io.Reader) (*simplejson.Json, error) {
+type wrappedResp struct {
+	Status     string      `json:"status_txt"`
+	StatusCode int         `json:"statuc_code"`
```
twmb: status_code not statuc_code

api_request.go
```diff
@@ -39,40 +38,52 @@ func newDeadlineTransport(timeout time.Duration) *http.Transport {
...
+	wResp := &wrappedResp{
+		Data: ret,
 	}
-	return data.Get("data"), nil
+
+	if err = json.Unmarshal(respBody, wResp); err != nil {
+		return err
+	}
+
+	// gResp.StatusCode here is equal to resp.StatusCode, so ignore it
```
twmb: wResp not gResp

api_request.go
twmb: I would just add a line of documentation that this function "stores the result in the value pointed to by ret" or something along those lines, the same way that json.Unmarshal is documented as "Unmarshal parses the JSON-encoded data and stores the result in the value pointed to by v."

That clarifies why the interface{} works in both cases.

twmb:
I couldn't find any guarantee that the lookupd response wouldn't change anywhere; maybe that should be documented inside nsq.

Other than my minor nits lgtm.

chzyer:
@twmb 
You are right. The `lookupd` response may change further, and I think that `lookupd` just return a `map` rather than a struct may make people make mistakes more easily while we changing the code.
Anyway, even `lookupd` changed, it will be backward compatibility.
btw, thanks for you review! I pushed a new commit :)

mreiferson:
:100: thanks :+1:

PR #140, opened 2015-05-08
Add a tcp no delay option

chzyer:
For a consumer to `nsqd`, each messages will sent `FIN` to `nsqd` if it processes successfully.  
I bench test consumer 10k message per second, the Ethernet is overhead by trigger too much interrupt requests to it(about 15k/s), and got very high(40%) hiq cpu usage. 
When I turn off the tcpNoDelay, the interrupts per second decrease to about 10k/s and the performance has improved.
ps: Loopback may not trigger high hiq cpu usage, but the performance is still improved.

### DIFF
--- a/config.go
+++ b/config.go
@@ -104,6 +104,10 @@ type Config struct {
 	ReadTimeout  time.Duration `opt:"read_timeout" min:"100ms" max:"5m" default:"60s"`
 	WriteTimeout time.Duration `opt:"write_timeout" min:"100ms" max:"5m" default:"1s"`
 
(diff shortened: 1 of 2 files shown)


mreiferson:
@chzyer thanks for digging into this a bit.  I'm concerned about enabling `TCP_NODELAY` by default and I'm not sure I want to rely on it to essentially buffer for us.

However, I think it _is_ worth thinking about library level write buffers (there are already read buffers) on network IO to increase throughput and reduce syscalls.

PR #142, opened 2015-05-16
consumer: backoff option max_backoff_level

hankya:
When a consumer failed to consume the message, the backoff_level count will keep increasing till message consumer is fixed, if the failure is long enough, the backoff_level count could too big, say 1000, then it will take like 1000*backoff_duration to recover the message consume speed for the consumer.

introduce config max_backoff_level so that the time on recover message consuming speed is configurable, i.e, if it is configured to 10, then recovery time is 10 * backoff_duration at most.

### DIFF
--- a/config.go
+++ b/config.go
@@ -544,7 +544,7 @@ func coerce(v interface{}, typ reflect.Type) (reflect.Value, error) {
 		v, err = coerceBackoffStrategy(v)
 	default:
 		v = nil
-		err = errors.New(fmt.Sprintf("invalid type %s", typ.String()))
+		err = fmt.Errorf("invalid type %s", typ.String())
 	}
 	return valueTypeCoerce(v, typ), err
 }
@@ -576,10 +576,8 @@ func coerceString(v interface{}) (string, error) {
 		return fmt.Sprintf("%d", v), nil
 	case float32, float64:
 		return fmt.Sprintf("%f", v), nil
-	default:
-		return fmt.Sprintf("%s", v), nil
 	}
-	return "", errors.New("invalid value type")
+	return fmt.Sprintf("%s", v), nil
 }
 
 func coerceDuration(v interface{}) (time.Duration, error) {
--- a/conn.go
+++ b/conn.go
@@ -465,7 +465,8 @@ func (c *Conn) auth(secret string) error {
 		return err
 	}
 
-	c.log(LogLevelInfo, "Auth accepted. Identity: %q %s Permissions: %d", resp.Identity, resp.IdentityUrl, resp.PermissionCount)
+	c.log(LogLevelInfo, "Auth accepted. Identity: %q %s Permissions: %d",
+		resp.Identity, resp.IdentityUrl, resp.PermissionCount)
 
 	return nil
 }
--- a/consumer.go
+++ b/consumer.go
@@ -335,7 +335,7 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
 	return nil
 }
 
-// ConnectToNSQLookupd adds multiple nsqlookupd address to the list for this Consumer instance.
+// ConnectToNSQLookupds adds multiple nsqlookupd address to the list for this Consumer instance.
 //
 // If adding the first address it initiates an HTTP request to discover nsqd
 // producers for the configured topic.
@@ -462,7 +462,7 @@ func (r *Consumer) queryLookupd() {
 		return
 	}
 
-	nsqdAddrs := make([]string, 0)
+	var nsqdAddrs []string
 	for _, producer := range data.Producers {
 		broadcastAddress := producer.BroadcastAddress
 		port := producer.TCPPort
@@ -482,7 +482,7 @@ func (r *Consumer) queryLookupd() {
 	}
 }
 
-// ConnectToNSQD takes multiple nsqd addresses to connect directly to.
+// ConnectToNSQDs takes multiple nsqd addresses to connect directly to.
 //
 // It is recommended to use ConnectToNSQLookupd so that topics are discovered
 // automatically.  This method is useful when you want to connect to local instance.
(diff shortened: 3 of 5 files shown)


jehiah:
@hankya have you evaluated setting the existing setting `MaxBackoffDuration` to a smaller value to bound your recovery time?

mreiferson:
right, as @jehiah pointed out - `MaxBackoffDuration` is _supposed_ to have a bounding effect - is there a bug?

hankya:
sorry i should have explained it more clear, the point is, before backoffLevel (the backoffCounter in below code in consumer.go) is decreasing to 0, consumer will only process 1 message and then continue the backoff timeout process.
So, if backoffLevel is 100, MaxBackoffDuration is 2m, it will take about 100 * 2 m to recover to the full speed (MaxInFlight count), one alternative way to fix this is to restart the consumer so the count is reset to 0, ut i think adding a config option is a better way.

```go
	if r.backoffCounter == 0 && backoffUpdated {
		// exit backoff
		count := r.perConnMaxInFlight()
		r.log(LogLevelWarning, "exiting backoff, returning all to RDY %d", count)
		for _, c := range r.conns() {
			r.updateRDY(c, count)
		}
	} else if r.backoffCounter > 0 {
		// start or continue backoff
		backoffDuration := r.config.BackoffStrategy.Calculate(int(backoffCounter))

		r.log(LogLevelWarning, "backing off for %.04f seconds (backoff level %d), setting all to RDY 0",
			backoffDuration.Seconds(), backoffCounter)

		// send RDY 0 immediately (to *all* connections)
		for _, c := range r.conns() {
			r.updateRDY(c, 0)
		}

		r.backoff(backoffDuration)
	}
```

so if

mreiferson:
I'm still confused - `MaxBackoffDuration` is the _total_ time the consumer should be allowed to backoff, not multiplied by failed attempt.

Here is the code that _should_ apply the aforementioned bounds:

https://github.com/bitly/go-nsq/blob/master/consumer.go#L778-L784

hankya:
@mreiferson, thank very much for your patience and sorry for my poor english:).

If you look at following logs, the last line indicate the consumer returned to the full speed when backoff level is 0, before that, the speed is always 1 (backoff timeout expired, sending RDY 1).

```
2015/05/19 11:33:28 WRN    1 [test/test] backing off for 32.0000 seconds (backoff level 5), setting all to RDY 0
2015/05/19 11:33:46 INF    1 [test/test] querying nsqlookupd http://192.168.1.6:4161/lookup?topic=test
2015/05/19 11:34:00 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:34:00 ERR    1 [test/test] Handler returned error (Post http://127.0.0.1:8005: dial tcp 127.0.0.1:8005: connection refused) for msg 08547057190ab000
2015/05/19 11:34:00 WRN    1 [test/test] backing off for 64.0000 seconds (backoff level 6), setting all to RDY 0
2015/05/19 11:34:46 INF    1 [test/test] querying nsqlookupd http://192.168.1.6:4161/lookup?topic=test
2015/05/19 11:35:04 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:35:04 WRN    1 [test/test] backing off for 32.0000 seconds (backoff level 5), setting all to RDY 0
2015/05/19 11:35:36 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:35:36 WRN    1 [test/test] backing off for 16.0000 seconds (backoff level 4), setting all to RDY 0
2015/05/19 11:35:46 INF    1 [test/test] querying nsqlookupd http://192.168.1.6:4161/lookup?topic=test
2015/05/19 11:35:52 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:35:52 WRN    1 [test/test] backing off for 8.0000 seconds (backoff level 3), setting all to RDY 0
2015/05/19 11:36:00 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:36:00 WRN    1 [test/test] backing off for 4.0000 seconds (backoff level 2), setting all to RDY 0
2015/05/19 11:36:04 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:36:04 WRN    1 [test/test] backing off for 2.0000 seconds (backoff level 1), setting all to RDY 0
2015/05/19 11:36:06 WRN    1 [test/test] (replace:4150) backoff timeout expired, sending RDY 1
2015/05/19 11:36:06 WRN    1 [test/test] exiting backoff, returning all to RDY 200
```

When there are two many failed attempts, it will result in a very big backoff level count, say 100, during the period of counting down backoff level from 100 to 0, the speed for consumer to be pushed for message is always 1, with this speed plus backoff , i can hardly consider it as working. I wonder why not limit the backoff level to reduce this period, then i can have control on the time for consumer to return to a health state when remote failure is fixed.

mreiferson:
@hankya don't worry your english is fine!

Those logs look fine to me - it is behaving as I expect.  What I'm pointing out is that backoff level _should_ be bounded (indirectly) via this code here:

https://github.com/bitly/go-nsq/blob/master/consumer.go#L778-L784

Can you show me the rest of the code you're using to configure the `nsq.Consumer`?

hankya:
Now I see,  and I found following code is causing the problem I am talking about.

https://github.com/bitly/go-nsq/blob/master/config.go#L48-50

It make sure that nextBackoff will always be less than MaxBackoffDuration, so the if condition in following highlighted code will always be true.

https://github.com/bitly/go-nsq/blob/master/consumer.go#L778-L784

Comment out https://github.com/bitly/go-nsq/blob/master/config.go#L48-50 will fix it.

P.S i was simple using nsq_to_http app as a consumer.

mreiferson:
@hankya ahhh, nice catch!

Would you like to update your patch on this PR here?

jehiah:
@hankya++ for persistence in figuring this out

judwhite:
is this still an issue for `FullJitterStrategy` ?

PR #150, opened 2015-07-15
update FullJitterStrategy Calculate to not bound by MaxBackoffDuration

judwhite:
- responsibility has been moved to Consumer startStopContinueBackoff
- see PR #147 and #142


mreiferson:
I was just about to say, it's not a "problem" but it is duplicate and unnecessary work.  Thanks for noticing/fixing!

judwhite:
I thought it would be a problem because of this https://github.com/bitly/go-nsq/blob/cfe4feafe8d2eccdd377c18eae70ba564f2d74ae/consumer.go#L780-L783

mreiferson:
I missed that!

PR #156, opened 2015-09-02
Add mutex around rng *rand.Rand to avoid data race on resume

jamesgroat:
Currently there is a data race when a consumer hits a backoff condition and then resumes. The resume via time.AfterFunc gets called from a different goroutine.

Note: this could also have been solved using one of the existing mutexes  (r.mtx or r.backoffMtx).

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -106,7 +106,8 @@ type Consumer struct {
 	channel string
 	config  Config
 
-	rng *rand.Rand
+	rngMtx sync.Mutex
+	rng    *rand.Rand
 
 	needRDYRedistributed int32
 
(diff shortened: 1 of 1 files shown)


consumer.go
```diff
@@ -1004,6 +1009,8 @@ func (r *Consumer) redistributeRDY() {
 		availableMaxInFlight = 1 - atomic.LoadInt64(&r.totalRdyCount)
 	}
 
+	r.rngMtx.Lock()
+	defer r.rngMtx.Unlock()
```
mreiferson: can we reduce the size of this critical path and put this lock around _just_ the call to `r.rng.Int()`?

mreiferson:
thanks for this, just one minor comment!

jamesgroat:
I've addressed the comment.

mreiferson:
LGTM, do you mind squashing down to one commit?

consumer.go
```diff
@@ -106,7 +106,8 @@ type Consumer struct {
 	channel string
 	config  Config
 
-	rng *rand.Rand
+	rngMtx sync.RWMutex
```
jehiah: `RWMutex` is designed for separate read and write locks. Since we only want global locking can you use `sync.Mutex` here?

jamesgroat:
Squashed and changed to using sync.Mutex.

PR #158, opened 2015-09-13
TestProducerConnection checks 127.0.0.1:4150 and helpfully suggests running nsdq

glycerine:
…if it is not found.

TestProducerConnection was failing out of the box due to referring to 127.0.0.2 instead of 127.0.01.

Once the ip address is fixed, then this new user had to guess that nsqd had to be started manually on 127.0.0.1:4150 in the background as well for the test to succeed. This patch checks all errors (two were being ignored) and on TestProducerConnection failure suggests the possible fix for the problem: that the user should start nsqd before running the test.

### DIFF
--- a/producer_test.go
+++ b/producer_test.go
@@ -40,18 +40,27 @@ func (h *ConsumerHandler) HandleMessage(message *Message) error {
 	return nil
 }
 
+// requires that nsdq be running on 127.0.0.1:4150
 func TestProducerConnection(t *testing.T) {
 	config := NewConfig()
-	laddr := "127.0.0.2"
+	laddr := "127.0.0.1"
 
-	config.LocalAddr, _ = net.ResolveTCPAddr("tcp", laddr+":0")
+	var err error
+	config.LocalAddr, err = net.ResolveTCPAddr("tcp", laddr+":0")
+	if err != nil {
+		t.Fatalf("could not ResolveTCPAddr '%s' - %s", laddr+":0", err)
+	}
 
-	w, _ := NewProducer("127.0.0.1:4150", config)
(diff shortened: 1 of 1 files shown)


mreiferson:
@glycerine thanks, tests should actually be run via the `test.sh` helper script in the root of the repo.

I've added a note about this to our `README.md` (see 8a3b2fa)

Gonna close.

glycerine:
Ok. Thanks @mreiferson for the guidance.  test.sh fails, but I'll open a separate PR for that (easy) fix. Leaving this closed.

$ ./test.sh 
starting nsqlookupd
  logging to /tmp/nsqlookupd.mrdDZvX
starting nsqd --data-path=/tmp --lookupd-tcp-address=127.0.0.1:4160 --tls-cert=./test/server.pem --tls-key=./test/server.key --tls-root-ca-file=./test/ca.pem
  logging to /tmp/nsqlookupd.ED8BApj
# github.com/nsqio/go-nsq
config_flag_test.go:6:2: cannot find package "github.com/bitly/go-nsq" in any of:
    /home/jaten/pkg/go1.5.1/src/github.com/bitly/go-nsq (from $GOROOT)
    /home/jaten/src/github.com/bitly/go-nsq (from $GOPATH)
FAIL    github.com/nsqio/go-nsq [setup failed]
killing nsqd PID 23809
killing nsqlookupd PID 23802
jaten@i7:~/src/github.com/nsqio/go-nsq:(0f97a46...)$

glycerine:
nevermind -- test.sh works fine on master, it is just when pinned to 0f97a46 per the gpm build that it fails

PR #162, opened 2015-09-29
producer: expose DeferredPublish command

DanielHeckrath:
Since a producer doesn't expose it's connection or send command
method, it might be beneficial to add helper functions for producer
related commands like DeferredPublish


mreiferson:
LGTM, thanks!

jackluo2012:
Is the problem solved by golang NSQ delay receiving message? I don't use this on this side. Can you give me an example

PR #163, opened 2015-10-01
TestProducerConnection() now green on OSX. fixes #161.

### LINKED ISSUE #161: producer_test.go:54: should lazily connect - dial tcp 127.0.0.2:0->127.0.0.1:4150: bind: can't assign requested address
I pulled to latest go-nsq, and find that when running test.sh, the TestProducerConnection fails. I'm on OSX 10.10.2, go 1.5.1, nsqd v0.3.6-alpha (built w/go1.5.1).

~~~
=== RUN   TestProducerConnection
--- FAIL: TestProducerConnection (0.00s)
        producer_test.go:54: should lazily connect - dial tcp 127.0.0.2:0->127.0.0.1:4150: bind: can't assign requested address
~~~

glycerine:
per discussion in https://github.com/nsqio/go-nsq/issues/161, this makes the TestProducerConnection() test (and now all tests) green on OSX.

### DIFF
--- a/producer_test.go
+++ b/producer_test.go
@@ -42,7 +42,7 @@ func (h *ConsumerHandler) HandleMessage(message *Message) error {
 
 func TestProducerConnection(t *testing.T) {
 	config := NewConfig()
-	laddr := "127.0.0.2"
+	laddr := "127.0.0.1"
 
 	config.LocalAddr, _ = net.ResolveTCPAddr("tcp", laddr+":0")
 


mreiferson:
@glycerine in order to actually verify the port, we need to update line 58 to use the whole local address, right now it's just `127.0.0.1`, right?

glycerine:
Good point. 

So I naively just now put in string comparison check. 

Arg. 

It turns out that the config.LocalAddr.String() is actually "127.0.0.1:0" -- with a zero port. We won't actually know what port is until it is bound. There is a race condition built in to the process of knowing a free port in advance, and telling the Producer to use it.  Ugh.

Elaboration: We could get a free port manually[1] in the test, but there's no guarantee that the port would still be free by the time the Publish() call tried to bind it. We would just be introducing a race condition into the test which would be rather unfortunate. It would succeed 99.99% of the time, and annoy the heck out of somebody that 0.01 % of the time by failing when run on a machine that has alot of tcp activity.

In trying to invent a work-around, I'm stuck. In a way I'm back to not really understanding the real intent of the test; I'm not sure what failure scenario is being guarded against. Do we really not trust that the parameters specified for LocalAddr will be used by the Producer?  That seems really unlikely. I can see wanting to be able to specify the use of a particular interface card, but such a test becomes non-portable very quickly. In sum, I can't think of a way to avoid the race condition and it doesn't seem worth the value received in exchange for adding a racy test. My judgement falls on this side: I would rather trust that the Publish() call is using the parameters it is specified to use, rather than introduce a racy test.

That said, in the past[1] I've certainly used the racy approach, and just kept a note to not worry if it flakes out. But that was for a personal project. For a continuously integrated and widely used piece of software, the medicine may be worse than the disease.

It's your call; if you want it, it's trivial and I'm happy to add code based on the approach in [1]. Let me know and I'll add it to the pull. Personally, I'm inclined to just forego that check. I'm a big fan of BDD and TDD, but everything is easily tested, or worth the tradeoffs.

[1] https://github.com/glycerine/goq/blob/master/ipaddr.go#L82

glycerine:
@mreiferson I've made a branch for each of the discussed approaches; pull as you prefer.

https://github.com/glycerine/go-nsq/tree/lighter-prodcheck
https://github.com/glycerine/go-nsq/tree/prespecify_port

mreiferson:
@glycerine thanks for digging in, again!

I realized `test.sh` _already_ requires NSQ's default ports to be open to work, so another option is to hard-code a port here rather than letting the OS pick one.

glycerine:
I imagine we could set an env variable in the script to pass the port into the test....

On balance though, I still favor actively grooming tests to maintain their value. I prefer to leave out parts of a test, or even entire tests, that don't carry their weight.  This helps future developers maintain a healthy respect for failing tests. That is, I suggest going with https://github.com/glycerine/go-nsq/tree/lighter-prodcheck

mreiferson:
> That is, I suggest going with https://github.com/glycerine/go-nsq/tree/lighter-prodcheck

Want squash in that last commit here and we can land it?

glycerine:
Squashed and pushed but this PR did not update. Opened https://github.com/nsqio/go-nsq/pull/164 in its place.

PR #176, opened 2016-04-22
consumer: delete nsqd addr when nsqd instance close if we connect to nsqlookup

FlamingTree:
as title


mreiferson:
What was the issue you were experiencing?

FlamingTree:
After rethink, it seem that it has no real affect on normal using the lib. so you can close the PR.

PR #178, opened 2016-05-06
expose cmd of struct ProducerTransaction

FlamingTree:
if PublishAsync return error, I need to know which Command cause error, so it's better expose cmd of struct ProducerTransaction.


mreiferson:
The idea is that you can use `PublishAsync` and pass `args`, which will be delivered over `doneChan` - e.g. you can use this to identify the command.

PR #194, opened 2016-09-29
Custom Broadcast address setter added

PumpkinSeed:
For remote connection it is better to set a custom broadcast address.


mreiferson:
The need for this indicates a misconfiguration with `nsqd`, so it isn't something that we should introduce workarounds for.

PumpkinSeed:
Okay, thank you for the review.

PR #201, opened 2016-12-01
fix producer Stop() EOF

mengskysama:
when producer.Stop() EOF exception occur

```

2016/12/01 09:44:47 ERR  3856 (10.160.3.70:4150) IO error - EOF
2016/12/01 09:44:47 INF  3856 (10.160.3.70:4150) beginning close
2016/12/01 09:44:47 INF  3856 (10.160.3.70:4150) readLoop exiting
2016/12/01 09:44:47 INF  3856 (10.160.3.70:4150) breaking out of writeLoop
2016/12/01 09:44:47 INF  3856 (10.160.3.70:4150) writeLoop exiting
2016/12/01 09:44:47 INF  3856 stopping
2016/12/01 09:44:47 INF  3856 exiting router

```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -479,7 +479,11 @@ func (c *Conn) readLoop() {
 		}
 
 		frameType, data, err := ReadUnpackedResponse(c)
+
 		if err != nil {
+			if err == io.EOF && atomic.LoadInt32(&c.closeFlag) == 1 {
+				goto exit
+			}
 			if !strings.Contains(err.Error(), "use of closed network connection") {
 				c.log(LogLevelError, "IO error - %s", err)
 				c.delegate.OnIOError(c, err)


conn.go
```diff
@@ -479,7 +479,11 @@ func (c *Conn) readLoop() {
...
 
 		frameType, data, err := ReadUnpackedResponse(c)
+
 		if err != nil {
+			if err == io.EOF && atomic.LoadInt32(&c.closeFlag) == 1 {
+				goto exit
+			}
 			if !strings.Contains(err.Error(), "use of closed network connection") {
 				c.log(LogLevelError, "IO error - %s", err)
 				c.delegate.OnIOError(c, err)
```
judwhite: @mreiferson @jehiah Should it check `c.closeFlag` if the error does contain "use of closed network connection"? Otherwise, it might skip a call to `c.delegate.OnIOError` which delegates to `(p *Producer) close()`. It's not immediately obviously what conditions inside the `net` package can lead to `net.errClosing`.

mreiferson: My assumption is that there's an edge case where you're cleanly shutting down but you _don't_ "use a closed connection" but instead hit an "expected" EOF.

This person was clearly getting a log message despite the existing logic, which I think confirms that.

PR #203, opened 2016-12-29
switch to golang/snappy

jehiah:
This matches nsqio/nsq#830


mreiferson:
Hah! good catch.

jehiah:
Also nsq Godeps needs to bump it's dependency of go-nsq to pick up this change; presumably it should have broken the build w/o this?

mreiferson:
> Also nsq Godeps needs to bump it's dependency of go-nsq to pick up this change; presumably it should have broken the build w/o this?

That's interesting, I can't explain how it succeeded...

PR #208, opened 2017-04-16
consumer: redistribute RDY when connections are active

### LINKED ISSUE #179: consumer: redistributeRDY fails when connection remains active
While experimenting with NSQ I've encountered a problem that looks like a flaw in the consumer code.

My setup is:
- There are two producers on different hosts.
- Each producer periodically (interval = 3 secs) posts to the same topic/channel to it's local `nsqd`.
- There is one consumer for this topic/channel. It discovers both `nsqd` nodes through `nsqlookupd`.
- The consumer uses the default configuration, and, thus, has `MaxInFlight` = 1.

It occurs, that with this setup the consumer is able to read only from _one nsqd of the two_. The second nsqd never gets a chance to send anything because consumer constantly keeps the correspondent RDY = 0 for the second nsqd connection.

I found that for such cases there is a special treatment in the code - every once in a while the consumer tries to randomly redistribute RDY among connected nsqd nodes (method `redistributeRDY`). However, in practice it never works because of this condition in the code:

``` go
availableMaxInFlight := int64(maxInFlight) - atomic.LoadInt64(&r.totalRdyCount)
...
for len(possibleConns) > 0 && availableMaxInFlight > 0 {
<redistribution logic>
}
```

Debugging shows that `availableMaxInFlight` always equal 0 at this moment, and actual redistribution never happens.

In fact, even if I hack a little, and force this redistribution to happen it does not work either, because there is a similar condition in method `updateRDY`. When there is a lack of available RDY, this method sets up a timer to retry in 5 seconds, and this attempt to retry always fails, and the timer is rescheduled again and again.

So, the questions are:
- Is it a bug or a feature?
- How this redistribution is supposed to work in such cases where active producers never give other producers a chance to receive even a minimal share of RDY/MaxInFlight?
- Am I supposed to be preventive when creating a consumer, and set up the value of MaxInFlight in such a way that it will most likely be bigger than a potential number of producing nsqd nodes? Does not seem as very viable solution in the long term.

mreiferson:
fixes #179

### DIFF
--- a/config.go
+++ b/config.go
@@ -125,10 +125,11 @@ type Config struct {
 	// Maximum number of times this consumer will attempt to process a message before giving up
 	MaxAttempts uint16 `opt:"max_attempts" min:"0" max:"65535" default:"5"`
 
-	// Duration to wait for a message from a producer when in a state where RDY
-	// counts are re-distributed (ie. max_in_flight < num_producers)
+	// Duration to wait for a message from an nsqd when in a state where RDY
+	// counts are re-distributed (e.g. max_in_flight < num_producers)
 	LowRdyIdleTimeout time.Duration `opt:"low_rdy_idle_timeout" min:"1s" max:"5m" default:"10s"`
-
+	// Duration to wait until redistributing RDY for an nsqd regardless of LowRdyIdleTimeout
+	LowRdyTimeout time.Duration `opt:"low_rdy_timeout" min:"1s" max:"5m" default:"30s"`
 	// Duration between redistributing max-in-flight to connections
 	RDYRedistributeInterval time.Duration `opt:"rdy_redistribute_interval" min:"1ms" max:"5s" default:"5s"`
 
--- a/conn.go
+++ b/conn.go
@@ -53,6 +53,7 @@ type Conn struct {
 	maxRdyCount      int64
 	rdyCount         int64
 	lastRdyCount     int64
+	lastRdyTimestamp int64
 	lastMsgTimestamp int64
 
 	mtx sync.Mutex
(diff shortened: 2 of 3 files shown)


mreiferson:
RFR @jehiah @judwhite @ploxiln 

This is one simple way to address #179 by bounding the time a consumer will stay pinned to any producing nsqd, regardless of last message delivery.

config.go
```diff
@@ -125,10 +125,11 @@ type Config struct {
...
 	MaxAttempts uint16 `opt:"max_attempts" min:"0" max:"65535" default:"5"`
 
-	// Duration to wait for a message from a producer when in a state where RDY
-	// counts are re-distributed (ie. max_in_flight < num_producers)
+	// Duration to wait for a message from an nsqd when in a state where RDY
+	// counts are re-distributed (e.g. max_in_flight < num_producers)
 	LowRdyIdleTimeout time.Duration `opt:"low_rdy_idle_timeout" min:"1s" max:"5m" default:"10s"`
-
+	// Duration to wait until redistributing RDY for an nsqd regardless of LowRdyIdleTimeout
+	LowRdyTimeout time.Duration `opt:"low_rdy_timeout" min:"1s" max:"5m" default:"30s"`
```
mreiferson: This should probably never be lower than `LowRdyIdleTimeout`, I can enforce this...

ploxiln:
Looks good to me.

I noticed this part below seems less than optimal (abbreviated):

```go
 	for len(possibleConns) > 0 && availableMaxInFlight > 0 {
 		availableMaxInFlight--
 		i := r.rng.Int() % len(possibleConns)
 		c := possibleConns[i]
 		possibleConns = append(possibleConns[:i], possibleConns[i+1:]...)
 		r.updateRDY(c, 1)
}
```

All `conns` are added to `possibleConns`, so it's quite possible that some of them already have non-zero ready count. So some of availableMaxInFlight will end up not actually distributed. (But this problem will probably fix itself in later redistributions. And it's not new.)

This could be fixed by appending to possibleConns only if the conn has a zero ready count.

mreiferson:
> All conns are added to possibleConns, so it's quite possible that some of them already have non-zero ready count. So some of availableMaxInFlight will end up not actually distributed.

Hmm, yea, can probably be improved, but it's at least randomized.

mreiferson:
@ploxiln want to gimme a 👍 — also, there are probably a few lessons to learn from nsqio/pynsq#179 and nsqio/pynsq#176

ploxiln:
👍 to these additions.

... but the existing rdyCount logic is a bit tricky, and I'm not really familiar with it, and it's taking a while to digest. Anyway, I'm confident that this is strictly an improvement. fwiw :)

PR #211, opened 2017-06-08
make query nsqlookupd dail timeout configurable

guyannanfei25:
when network is not well, 2 seconds may not be able dail success.
Use nsq.Config.DialTimeout  to configure this parameter

### DIFF
--- a/api_request.go
+++ b/api_request.go
@@ -46,8 +46,8 @@ type wrappedResp struct {
 }
 
 // stores the result in the value pointed to by ret(must be a pointer)
-func apiRequestNegotiateV1(method string, endpoint string, body io.Reader, ret interface{}) error {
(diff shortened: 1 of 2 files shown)


mreiferson:
I think this is a good change, thanks for the contribution! However, I don't think we should lean on the existing `DialTimeout`.

Can we add new `LookupdDialTimeout` and `LookupdRequestTimeout` and thread those into the transport instantiation?

guyannanfei25:
I think `DialTimeout` means parameter for all connections. Considering lookupds and nsqds are usually in same idc, the network condition is same. Besides even not in same idc, `DialTimeout` sets a little bigger do no harm for the ones in good network condition. Do we need new `LookupdDialTimeout`  and `LookupdRequestTimeout`  parameters ?

mreiferson:
My thinking is that if we're going to improve this, might as well do it completely and in an obvious way.

Issue #171, opened 2016-03-25
crash in (*Producer).popTransaction


chzyer:
Hi, guys. I have meet a unexpected panic in producer, but I couldn't figure out how it happens.

Here is the panic log, and the version of `go-nsq` is newest.
```
panic: runtime error: index out of range

goroutine [running]:
runtime.panic(0x7b5a60, 0xa4457c)
	/usr/local/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/nsqio/go-nsq.(*Producer).popTransaction()
	/data/www/go_proj/src/github.com/nsqio/go-nsq/producer.go:336 +0x199
github.com/nsqio/go-nsq.(*Producer).router()
	/data/www/go_proj/src/github.com/nsqio/go-nsq/producer.go:319 +0x3dd
created by github.com/nsqio/go-nsq.(*Producer).connect
	/data/www/go_proj/src/github.com/nsqio/go-nsq/producer.go:290 +0x651
```

I check `(*Producer) popTransaction()`, `t := w.transactions[0]` cause the crash, so I check all the functions which manipulate `w.transactions`. Unfortunately, all the functions only use in `(*Producer).router()`, which means there is not possible occurs the data races things. 

So, Is there any other possible ? Thanks.

jehiah:
definitely sounds like a bug of some sort. @chzyer can you confirm what version go-nsq you are using, and w/ what version of Go.

Also, to clarify was this a panic you saw once, or is it recurring for you?

jehiah:
The only thing i can think of is if a connection gets an unexpected `FrameTypeError` that doesn't match a response, or if there is a case where both a response and error are possible. thoughts @mreiferson ?

mreiferson:
Strange, never seen this before.  @chzyer do you have more of the logs for the process that crashed so we can see the event in context?

RobitYadda:
I meet this too. It throw error massage "not connect" before panic.

jehiah:
@RobitYadda can you share logs that include the context of your error?

RobitYadda:
My code like this:
for{
//other process
	go func() {
		err = writer.Publish("events", msgBytes)
		if err != nil {
			glog.Errorf("%v", err)
		}
	}()
}

Logs(some are logged nsq-client,others are logged by glog above):

2016/05/06 12:43:30 ERR    1 (127.0.0.1:4150) IO error - write tcp 127.0.0.1:40915->127.0.0.1:4150: write: broken pipe
[...]

PR #217, opened 2017-09-05
feat(producer): avoid sending duplicate data

zwb-ict:
``` go
for {
	select {
	case t := <-w.transactionChan:
		w.transactions = append(w.transactions, t)
		err := w.conn.WriteCommand(t.cmd)
		if err != nil {
			w.log(LogLevelError, "(%s) sending command - %s", w.conn.String(), err)
			w.close()
		}
	case data := <-w.responseChan:
		w.popTransaction(FrameTypeResponse, data)
	case data := <-w.errorChan:
		w.popTransaction(FrameTypeError, data)
	case <-w.closeChan:
		goto exit
	case <-w.exitChan:
		goto exit
	}
}
```

after producer called Stop, there may be data in channal w.responseChan,  and exitChan may has been closed, and select randomly selected goto exit. 

then in 

``` go
func (w *Producer) transactionCleanup() {
	// clean up transactions we can easily account for
	for _, t := range w.transactions {
		t.Error = ErrNotConnected
		t.finish()
	}
	w.transactions = w.transactions[:0]
	... ...
}
```
if we don't read responseChan, the successfully sent data may be send again after creating a new producer and publish again.


mreiferson:
@zwb-ict great catch! I think this might explain the long-standing bug in #171! ⭐️

producer.go
```diff
@@ -342,12 +342,21 @@ func (w *Producer) popTransaction(frameType int32, data []byte) {
...
 
 func (w *Producer) transactionCleanup() {
-	// clean up transactions we can easily account for
-	for _, t := range w.transactions {
-		t.Error = ErrNotConnected
-		t.finish()
+l:
+	for {
+		select {
+		case data := <-w.responseChan:
```
mreiferson: I think we probably have the same race in `w.errorChan`, so we should flush that channel too.

producer.go
```diff
@@ -342,12 +342,21 @@ func (w *Producer) popTransaction(frameType int32, data []byte) {
 }
 
 func (w *Producer) transactionCleanup() {
-	// clean up transactions we can easily account for
-	for _, t := range w.transactions {
-		t.Error = ErrNotConnected
-		t.finish()
+l:
```
mreiferson: I don't _think_ this label is necessary since it's not a nested `for` loop.

producer.go
```diff
@@ -342,12 +342,21 @@ func (w *Producer) popTransaction(frameType int32, data []byte) {
...
+	for {
+		select {
+		case data := <-w.responseChan:
+			w.popTransaction(FrameTypeResponse, data)
```
mreiferson: Should these be popped and finished with `ErrNotConnected`, like below?

producer.go
zwb-ict: In ```for``` loop, there's a ```select``` in it, without label, ```break``` only jump out from ```select``` not ```for```.

zwb-ict: In ```for``` loop, there's a ```select``` in it, without label, ```break``` only jump out from ```select``` not ```for```.

producer.go
zwb-ict: I think you're right. The **real error reason** should pass to user, not just set  **error reason** to ```ErrNotConnected```. I will add this code.

producer.go
zwb-ict: I think the release code like below：


``` go

func (w *Producer) transactionCleanup() {
l:
	for {
		select {
		case data := <-w.responseChan:
			w.popTransaction(FrameTypeResponse, data)
		case data := <-w.errorChan:
			w.popTransaction(FrameTypeError, data)
		default:
			// clean up transactions we can easily account for
			for _, t := range w.transactions {
				t.Error = ErrNotConnected
				t.finish()
			}
			w.transactions = w.transactions[:0]
			break l
		}
	}
        ... ...
}
```

The  transaction poped in ```chan w.responseChan``` branch means successfully sent, so transaction's error reason is empty and should not be set.
The  transaction poped in ```chan w.errorChan``` branch means sent failed, should be send again, and transaction's error reason should be set to the real reason, so user can know what happened.

zwb-ict:
@mreiferson 
Is there a possible when user called ```Stop```? 

``` go
func (w *Producer) Stop() {
	w.guard.Lock()
	if !atomic.CompareAndSwapInt32(&w.stopFlag, 0, 1) {
		w.guard.Unlock()
		return
	}
	w.log(LogLevelInfo, "stopping")
	close(w.exitChan)
	w.close()
	w.guard.Unlock()
	w.wg.Wait()
}
```

```close(w.exitChan)``` is called before ```w.close()``` (which will call w.conn.close())

So router may exit ```for``` because of  ```case <-w.exitChan:```

``` go
func (w *Producer) router() {
	for {
		select {
		case t := <-w.transactionChan:
			w.transactions = append(w.transactions, t)
			err := w.conn.WriteCommand(t.cmd)
			if err != nil {
				w.log(LogLevelError, "(%s) sending command - %s", w.conn.String(), err)
				w.close()
			}
		case data := <-w.responseChan:
			w.popTransaction(FrameTypeResponse, data)
		case data := <-w.errorChan:
			w.popTransaction(FrameTypeError, data)
		case <-w.closeChan:
			goto exit
		case <-w.exitChan:
			goto exit
		}
	}

exit:
	w.transactionCleanup()
	w.wg.Done()
	w.log(LogLevelInfo, "exiting router")
}
```

Before producer'goroutine run  ```w.transactionCleanup()```.  conn's goroutine may pass data into ```w.responseChan``` or ```w.errorChan``` before ```w.conn.close()```.

zwb-ict:
@mreiferson 
below https://github.com/nsqio/go-nsq/blob/master/conn.go#L536, should we add `goto exit`
like https://github.com/nsqio/go-nsq/blob/master/conn.go#L508?

mreiferson:
> Is there a possible when user called Stop?

@zwb-ict yes, it looks like in that case the race would be possible.

> below https://github.com/nsqio/go-nsq/blob/master/conn.go#L536, should we add goto exit like https://github.com/nsqio/go-nsq/blob/master/conn.go#L508?

Hmmm, possibly, but let's not do that in this PR?

zwb-ict:
@mreiferson yes, let's not do that in this PR.

PR #222, opened 2018-01-08
consumer: update RDY when setting MaxInFlight to 0 (pause consumer)

mscso:
When trying to pause the consumer via MaxInFlight 0, updateRDY is not called as maybeUpdateRDY checks for "count > 0". This PR removes that check from the condition.

Now I'm not sure if there's a legitimate reason for not updating RDY when MaxInFlight is 0, but right now the code breaks the "pause consumer" use case.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -884,7 +884,7 @@ func (r *Consumer) maybeUpdateRDY(conn *Conn) {
 	count := r.perConnMaxInFlight()
 
 	// refill when at 1, or at 25%, or if connections have changed and we're imbalanced
-	if remain <= 1 || remain < (lastRdyCount/4) || (count > 0 && count < remain) {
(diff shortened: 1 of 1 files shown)


mreiferson:
Good catch! I think we need to approach this a little differently, perhaps by special casing `0` in `ChangeMaxInFlight()` and making sure to check/clear out any backoff state?

mscso:
Not sure which special casing you mean - I can see the comment "For example, ChangeMaxInFlight(0) would pause message flow" - but no actual check for it?

mreiferson:
Exactly, I think `ChangeMaxInFlight` should bypass normal RDY handling and set connections to 0, but it needs to also account for backoff state on any of those connections.

mscso:
Ah understood and I guess that makes sense. Personally I'd chose to not add additional code paths and simplify instead - but that's your call :)

Let me know if I can help.

mreiferson:
gonna fix this in #222, you ended up being on the right track after all, thanks!

PR #224, opened 2018-01-21
make getLogger public

andyxning:
This PR will make `getLogger` public. 

For now, 
* there is no way to get currently used logger instance
* there is no way to just set the log level

With `getLogger` being public, we can easily resolve the above two problems.
> there is no way to get currently used logger instance

`GetLogger()` will return log instance and log level

> there is no way to just set the log level
```
logger, _ := consumer.GetLogger()
consumer.SetLogger(logger, LogLevelDebug)

```

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -223,7 +223,7 @@ func (r *Consumer) SetLogger(l logger, lvl LogLevel) {
 	r.logLvl = lvl
 }
 
-func (r *Consumer) getLogger() (logger, LogLevel) {
+func (r *Consumer) GetLogger() (logger, LogLevel) {
 	r.logGuard.RLock()
 	defer r.logGuard.RUnlock()
 
@@ -523,7 +523,7 @@ func (r *Consumer) ConnectToNSQD(addr string) error {
 
 	atomic.StoreInt32(&r.connectedFlag, 1)
 
-	logger, logLvl := r.getLogger()
+	logger, logLvl := r.GetLogger()
 
 	conn := NewConn(addr, &r.config, &consumerConnDelegate{r})
 	conn.SetLogger(logger, logLvl,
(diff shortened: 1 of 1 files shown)


mreiferson:
I'm -1 on this. The default logger is from the stdlib (https://github.com/nsqio/go-nsq/blob/master/consumer.go#L169) and if you want to change the log level you can just pass in your own instantiation of it.

If we want to more thoughtfully and broadly rethink logging for go-nsq in light of evolving Go library best practices I'm all for it.

andyxning:
> The default logger is from the stdlib (https://github.com/nsqio/go-nsq/blob/master/consumer.go#L169) and if you want to change the log level you can just pass in your own instantiation of it.

Without reading the source code, i do not know that the default logger is from `log`. This is fine to developers, but as for normal users we should have a simple way to just change the log level with original default logger. But as for now there is no way to do this. :)

Otherwise, how about we add a function to set the loglevel `SetLogLevel`.

ploxiln:
Maybe what this calls for is some more documentation, like adding an example of setting a log level with a default-type logger here: https://godoc.org/github.com/nsqio/go-nsq#Consumer.SetLogger

andyxning:
> Maybe what this calls for is some more documentation, like adding an example of setting a log level with a default-type logger

@ploxiln Thanks for the suggestion. But i still think we should make users available to get the currently used logger. Or on the other hand why we do not expose the currently used logger? This is a common requirements. :)

andyxning:
@ploxiln @mreiferson Any updates?

mreiferson:
@andyxning sorry for the delay! I agree with @ploxiln's suggestion:

> Maybe what this calls for is some more documentation, like adding an example of setting a log level with a default-type logger here: https://godoc.org/github.com/nsqio/go-nsq#Consumer.SetLogger

andyxning:
@mreiferson Seems good to me. I will file another PR to do this. :)

mreiferson:
Thanks!

PR #227, opened 2018-04-28
consumer: order-of-operations for messagesReceived

qianlnk:
1、messagesReceived should be add after message had transport to message channel.

2、 add Consume method,use Consume to get message channel if someone want to custom  `ack` or `nack`.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -198,6 +198,11 @@ func (r *Consumer) Stats() *ConsumerStats {
 	}
 }
 
+//Consume return message chan, get message without handler
+func (r *Consumer) Consume() chan *Message {
+	return r.incomingMessages
+}
+
 func (r *Consumer) conns() []*Conn {
 	r.mtx.RLock()
 	conns := make([]*Conn, 0, len(r.connections))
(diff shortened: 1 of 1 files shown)


mreiferson:
@qianlnk thanks for the PR, but I don't actually understand the bug. Also, for numerous reasons, it's not safe to expose the go channel over which messages are delivered, so we wouldn't merge that API addition.

qianlnk:
@mreiferson We can't receive the message from message channel until the message transfer to message channel(because the message channel may be blocked). I hold the opinion  that count of `messagesReceived` should be put after the message transfer to message channel. as for  the user, `messagesReceived` must be equal to the number received from message handler.

ploxiln:
You propose to not increment `messagesReceived` until the handler starts processing the message. I think the original intention is to count the message received as soon as the process receives it from nsqd (and I think this is even more important for `totalRdyCount`).

qianlnk:
ok, I know what you mean. But when I run `consume.Stop` and get `ConsumerStats` by `Stats` method，I find that `stats.MessagesReceived` != `stats.MessagesFinished` + `stats.MessagesRequeued` sometime. if `messagesReceived` increase here,  `messagesRequeued` should also increase when the `Stop`?

ploxiln:
hmm ... I think it might be possible that some messages could be received by the consumer process from nsqd, just after Stop() is called and `incomingMessages` is closed, and we could possibly handle it more elegantly, though it's a tricky distributed timing problem ... what you could do is call [ChangeMaxInFlight(0)](https://godoc.org/github.com/nsqio/go-nsq#Consumer.ChangeMaxInFlight) and wait a couple seconds, then call `Stop()`, and in that case you should get consistent numbers, and no dropped messages

PR #239, opened 2018-10-28
tests: update travis-ci test matrix

HaraldNordgren:


.travis.yml
```diff
@@ -2,7 +2,8 @@ language: go
 go:
   - 1.8.x
   - 1.9.x
-  - 1.10.x  
+  - 1.10.x
+  - 1.11.x
 env:
   - NSQ_DOWNLOAD=nsq-0.3.7.linux-amd64.go1.6 GOARCH=amd64
   - NSQ_DOWNLOAD=nsq-0.3.7.linux-amd64.go1.6 GOARCH=386
```
ploxiln: you want to remove nsq-0.3.7 and add nsq-1.1.0 ?

HaraldNordgren: You want to bump all these versions to 1.1.0?

```yaml
  - NSQ_DOWNLOAD=nsq-0.3.7.linux-amd64.go1.6 GOARCH=amd64
  - NSQ_DOWNLOAD=nsq-0.3.7.linux-amd64.go1.6 GOARCH=386
  - NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2 GOARCH=amd64
  - NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2 GOARCH=386
  - NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8 GOARCH=amd64
  - NSQ_DOWNLOAD=nsq-1.0.0-compat.linux-amd64.go1.8 GOARCH=386
```

ploxiln: No. Each release-build is only available compiled with a single go version, which happens to be embedded in the download filename. So just remove the two lines with nsq-0.3.7-..... and if you look at the nsq releases you can see that the release-build download-filename for nsq-1.1.0 is "nsq-1.1.0.linux-amd64.go1.10.3" so you can add two lines at the end for that one (one for each GOARCH)

PR #249, opened 2019-03-07
consumer: cleanup RDY handling; fix pausing

mreiferson:
this addresses the issue reported in #222

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -52,7 +52,6 @@ type Conn struct {
 	messagesInFlight int64
 	maxRdyCount      int64
 	rdyCount         int64
-	lastRdyCount     int64
 	lastRdyTimestamp int64
 	lastMsgTimestamp int64
 
@@ -207,13 +206,12 @@ func (c *Conn) RDY() int64 {
 
 // LastRDY returns the previously set RDY count
 func (c *Conn) LastRDY() int64 {
-	return atomic.LoadInt64(&c.lastRdyCount)
+	return atomic.LoadInt64(&c.rdyCount)
 }
 
 // SetRDY stores the specified RDY count
 func (c *Conn) SetRDY(rdy int64) {
 	atomic.StoreInt64(&c.rdyCount, rdy)
-	atomic.StoreInt64(&c.lastRdyCount, rdy)
 	if rdy > 0 {
 		atomic.StoreInt64(&c.lastRdyTimestamp, time.Now().UnixNano())
 	}
@@ -225,6 +223,8 @@ func (c *Conn) MaxRDY() int64 {
 	return c.maxRdyCount
 }
 
+// LastRdyTime returns the time of the last non-zero RDY
+// update for this connection
 func (c *Conn) LastRdyTime() time.Time {
 	return time.Unix(0, atomic.LoadInt64(&c.lastRdyTimestamp))
 }
@@ -523,7 +523,6 @@ func (c *Conn) readLoop() {
 			msg.Delegate = delegate
 			msg.NSQDAddress = c.String()
 
-			atomic.AddInt64(&c.rdyCount, -1)
 			atomic.AddInt64(&c.messagesInFlight, 1)
 			atomic.StoreInt64(&c.lastMsgTimestamp, time.Now().UnixNano())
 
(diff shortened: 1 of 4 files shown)


mreiferson:
RFR

mock_test.go
```diff
@@ -111,25 +112,25 @@ func (n *mockNSQD) handle(conn net.Conn) {
...
+			n.t.Logf("mock: %s", line)
 			n.got = append(n.got, line)
 			params := bytes.Split(line, []byte(" "))
 			switch {
 			case bytes.Equal(params[0], []byte("IDENTIFY")):
 				l := make([]byte, 4)
 				_, err := io.ReadFull(rdr, l)
 				if err != nil {
-					log.Printf(err.Error())
+					n.t.Logf(err.Error())
```
ploxiln: this should use `Log()` instead of `Logf()` (and 2 other cases below)

ploxiln:
I noticed that `rdyRetryMtx` and `backoffMtx` can be plain Mutex instead of RWMutex

ploxiln:
This is honestly the part of NSQ that I'm least familiar with ... the technical details of how ready-count is "refilled" and such ... I'll keep looking and slowly digesting though :)

mock_test.go
mreiferson: OK, fine

mreiferson:
Addressed comments.

> the technical details of how ready-count is "refilled" and such

Yea, this is just old and outdated code that's already been [addressed in `pynsq`](https://github.com/nsqio/pynsq/pull/179). See https://github.com/nsqio/nsq/pull/404 for relevant context.

mock_test.go
```diff
@@ -470,3 +469,73 @@ func TestConsumerBackoffDisconnect(t *testing.T) {
...
+	q.SetLogger(newTestLogger(t), LogLevelDebug)
+	q.AddHandler(&testHandler{})
+	err := q.ConnectToNSQD(n.tcpAddr.String())
+	if err != nil {
+		t.Fatalf(err.Error())
+	}
+
+	timeoutCh := time.After(500 * time.Millisecond)
+	pauseCh := time.After(50 * time.Millisecond)
+	unpauseCh := time.After(75 * time.Millisecond)
```
mreiferson: This is a little fragile and time dependent, but should be reasonably stable.

conn.go
```diff
@@ -207,13 +206,12 @@ func (c *Conn) RDY() int64 {
 
 // LastRDY returns the previously set RDY count
```
ploxiln: I guess this function is somewhat deprecated now?

mock_test.go
```diff
@@ -470,3 +469,73 @@ func TestConsumerBackoffDisconnect(t *testing.T) {
 		}
 	}
 }
+
+func TestConsumerPause(t *testing.T) {
+	msgIDGood := MessageID{'1', '2', '3', '4', '5', '6', '7', '8', '9', '0', 'a', 's', 'd', 'f', 'g', 'h'}
```
ploxiln: not an expected message id, as it has characters that are not hex ... not sure if it matters at all

very minor idea, alternate way to write this:

```go
	var msgId MessageID
	copy(msgId[:], []byte("0123456789abcdef"))
```

mock_test.go
```diff
@@ -470,3 +469,73 @@ func TestConsumerBackoffDisconnect(t *testing.T) {
...
+
+	msgGood := NewMessage(msgIDGood, []byte("good"))
+
+	script := []instruction{
+		// SUB
+		instruction{0, FrameTypeResponse, []byte("OK")},
+		// IDENTIFY
```
ploxiln: I think IDENTIFY is before SUB?

ploxiln:
looks good 👍

conn.go
mreiferson: Yea

mock_test.go
mreiferson: copied from previous tests and yes, it doesn't matter

mock_test.go
mreiferson: Fixing

PR #255, opened 2019-03-28
TLS Config copy fix

### LINKED ISSUE #254: go-vet: assignment copies lock value of tls.Config
While there are some minor linting issues, I noticed go-vet pointing out

>assignment copies lock value to conf: crypto/tls.Config contains sync.Once contains sync.Mutex

Here: https://github.com/nsqio/go-nsq/blob/61f49c096d0d767be61e4de06f724e1cb5fd85d7/conn.go#L390

Since the tls.Config is an option in the nsq.Config, you can't be certain it isn't already being used, and therefore synchronization of the Once/Mutex are lost in the copied value. Depending on the minimum Go version this package supports, the solution might be as simple as Cloning the tls.Config with the provided method added in Go1.8 https://golang.org/pkg/crypto/tls/#Config.Clone

iaburton:
Fixes #254 

I also ended up moving the net.SplitHostPort code up so if it failed we don't make a new tls.Config/clone for nothing.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -384,18 +384,19 @@ func (c *Conn) identify() (*IdentifyResponse, error) {
 }
 
 func (c *Conn) upgradeTLS(tlsConf *tls.Config) error {
-	// create a local copy of the config to set ServerName for this connection
-	var conf tls.Config
-	if tlsConf != nil {
-		conf = *tlsConf
-	}
 	host, _, err := net.SplitHostPort(c.addr)
 	if err != nil {
 		return err
 	}
+
+	// create a local copy of the config to set ServerName for this connection
+	conf := &tls.Config{}
+	if tlsConf != nil {
+		conf = tlsConf.Clone()
+	}
 	conf.ServerName = host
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -384,18 +384,19 @@ func (c *Conn) identify() (*IdentifyResponse, error) {
...
-	if tlsConf != nil {
-		conf = *tlsConf
-	}
 	host, _, err := net.SplitHostPort(c.addr)
 	if err != nil {
 		return err
 	}
+
+	// create a local copy of the config to set ServerName for this connection
+	conf := new(tls.Config)
```
mreiferson: minor nit: the convention we've followed for cases like this is `conf := &tls.Config{}`

iaburton: Personally I prefer `new` unless I'm also instantiating the types fields at the same time. 

Went ahead and changed it.

ploxiln:
one last thing: can you squash down to 1 commit?

PR #267, opened 2019-09-22
check response message size is positive

andyxning:
This PR adds pre check for msg size before `make` a slice. 

We have encountered a panic in `make([]byte, msgSize)` with `panic: runtime error: makeslice: len out of range`. We have bounded the max message size by 20MB. So, it seems that make function should not panic. 

I have test that when `make` with a negative length, `make` will panic with `panic: runtime error: makeslice: len out of range`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 go:
-  - 1.10.x
   - 1.11.x
   - 1.12.x
+  - 1.13.x
 env:
   - NSQ_DOWNLOAD=nsq-0.3.8.linux-amd64.go1.6.2 GOARCH=amd64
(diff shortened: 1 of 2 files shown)


ploxiln:
This change looks obviously correct.

I think the commit title could be more specific - this is for reading a reply message. Perhaps "check reply message size is positive"

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 go:
-  - 1.10.x
   - 1.11.x
   - 1.12.x
+  - 1.13.x
```
ploxiln: this is sort of throwing in an unrelated change ... but also seems fine

PR #275, opened 2019-11-23
Added support for setting logger for each log level

crazyweave:
I’ve added support for using separate loggers for each level.  I tried to minimize the changes required to existing code and maintain as much backwards compatibility as I could.

The change required to existing code is the requirement to add a call to `SetLoggingLevel()` if the default level (`LogLevelInfo`) is not the desired level. Previously the call to `SetLogger()` would do this implicitly.

This change also expands the API surface by 1 with a convenience function `SetLoggerForLevels()` that sets the same logger for the levels provided as parameters. This convenience function simplifies achieving the previous behavior of using just one logger for all levels.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -12,6 +12,8 @@ import (
 type producerConn interface {
 	String() string
 	SetLogger(logger, LogLevel, string)
+	SetLoggerLevel(LogLevel)
+	SetLoggerForLevel(logger, LogLevel, string)
 	Connect() (*IdentifyResponse, error)
 	Close() error
 	WriteCommand(*Command) error
@@ -28,7 +30,7 @@ type Producer struct {
 	conn   producerConn
 	config Config
 
-	logger   logger
+	logger   []logger
 	logLvl   LogLevel
 	logGuard sync.RWMutex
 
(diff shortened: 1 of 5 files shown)


jehiah:
@crazyweave can you describe what your use case is for this change?

crazyweave:
I was debugging the unusual number of high priority tickets generated from go-nsq and determined that instead of only error logs generating tickets, every log message being logged (given a logging level) generates a ticket.

The root of our issue was an incorrect assumption regarding the behavior of `SetLogger()` - only one logger and one logging level is allowed for the entire package.

This PR enables package consumers to employ a more nuanced approach to setting up logging for go-nsq.

This allows us to redirect error log messages to get the right attention while routing all other log messages elsewhere - including suppressing them by not setting a logger for that log level.

ploxiln:
I think that, if we are to go forward with this, we should keep more compatibility for existing public methods. `SetLogger()` should set the logger for all levels in order to be analogous to what it did before, and then you could have a new `SetLoggerForLevel()` to set one for a single level.

crazyweave:
@ploxiln Sounds perfectly reasonable.

crazyweave:
I've pulled the latest changes from master and made the requested changes.

crazyweave:
@jehiah or @ploxiln

producer.go
```diff
@@ -369,7 +402,8 @@ func (w *Producer) transactionCleanup() {
 }
 
 func (w *Producer) log(lvl LogLevel, line string, args ...interface{}) {
-	logger, logLvl := w.getLogger()
+	logger := w.getLogger(lvl)
+	logLvl := w.getLogLevel()
```
ploxiln: calling two methods here ends up doing RLock and RUnlock twice ... could still be a single function `getLogger()` that also returns current level

crazyweave: I prefer the separation but you make a good point.

crazyweave:
@ploxiln Thanks for reviewing this. Do you see anything else or have any more concerns?

ploxiln:
Nope, I don't see anything else, looks fine to me. I'm not really one of primary authors of this component, so I'll give @jehiah a bit more time to comment.

One last little thing though, can you squash down to one commit? Thanks :)

crazyweave:
@ploxiln - squashed to single commit.
@jehiah - Any concerns with this PR?

PR #283, opened 2020-01-16
remove duplicate 'Config.assertInitialized()' check

SwanSpouse:
```c.assertInitialized()```  is executed twice both in  ``` NewProducer() / NewConsumer() ```  and ```config.Validate()```


ploxiln:
Thanks - and sorry about taking a while to review.

PR #296, opened 2020-05-07
feature: nsqd connection over proxy (socks...)

HaoweiCh:
import "golang.org/x/net/proxy"

	if dialer, err := proxy.SOCKS5("tcp", "192.168.3.1:23456", nil, nil); err != nil {
		log.Fatal(err)
	} else {
		cfg.NSQ.Config.Dialer = dialer
	}

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,6 +36,12 @@ type BackoffStrategy interface {
 	Calculate(attempt int) time.Duration
 }
 
+// A Dialer is a means to establish a connection.
+type Dialer interface {
+	// Dial connects to the given address via the proxy.
+	Dial(network, addr string) (c net.Conn, err error)
+}
+
 // ExponentialStrategy implements an exponential backoff strategy (default)
 type ExponentialStrategy struct {
 	cfg *Config
@@ -92,6 +98,7 @@ type Config struct {
 	// used to Initialize, Validate
 	configHandlers []configHandler
 
+	Dialer      Dialer        `opt:"dialer"`
 	DialTimeout time.Duration `opt:"dial_timeout" default:"1s"`
 
 	// Deadlines for network reads and writes
--- a/conn.go
+++ b/conn.go
@@ -140,9 +140,12 @@ func (c *Conn) getLogger() (logger, LogLevel, string) {
 // Connect dials and bootstraps the nsqd connection
 // (including IDENTIFY) and returns the IdentifyResponse
 func (c *Conn) Connect() (*IdentifyResponse, error) {
-	dialer := &net.Dialer{
-		LocalAddr: c.config.LocalAddr,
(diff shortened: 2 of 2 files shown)


config.go
```diff
@@ -92,6 +98,7 @@ type Config struct {
 	// used to Initialize, Validate
 	configHandlers []configHandler
 
+	Dialer      Dialer        `opt:"dialer"`
 	DialTimeout time.Duration `opt:"dial_timeout" default:"1s"`
```
ploxiln, line 102: maybe these should be moved next to `LocalAddr` (just below) and get a little doc comment?

gknw:
I want to vote for accept this PR. We are looking for a way to manage NSQ connection, and a custom dialer solved (partially) our problem. It's a convenient thing.

config.go
```diff
@@ -101,6 +107,10 @@ type Config struct {
 	// LocalAddr is the local address to use when dialing an nsqd.
 	// If empty, a local address is automatically chosen.
 	LocalAddr net.Addr `opt:"local_addr"`
+	// Dialer affect connection when dialing an nsqd. Overwrite this to connect over proxy.
+	//
+	// Conflict with options LocalAddr and DialTimeout.
+	Dialer Dialer `opt:"dialer"`
```
mreiferson, line 113: We need to decide how to handle configuration specified for this `opt`, e.g. for command line utilities in `nsqd` that use `--consumer-opt`.

See: https://github.com/nsqio/nsq/blob/master/apps/nsq_to_file/nsq_to_file.go#L131-L135

And an example of how we handle `tls_config`: https://github.com/nsqio/go-nsq/blob/master/config.go#L388-L475

config.go
```diff
@@ -101,6 +107,10 @@ type Config struct {
 	// LocalAddr is the local address to use when dialing an nsqd.
 	// If empty, a local address is automatically chosen.
 	LocalAddr net.Addr `opt:"local_addr"`
+	// Dialer affect connection when dialing an nsqd. Overwrite this to connect over proxy.
+	//
+	// Conflict with options LocalAddr and DialTimeout.
```
mreiferson, line 112: Correct, this does conflict. We should probably error when a `dialer` and either of those two config fields are set.

config.go
```diff
@@ -101,6 +107,10 @@ type Config struct {
 	// LocalAddr is the local address to use when dialing an nsqd.
 	// If empty, a local address is automatically chosen.
 	LocalAddr net.Addr `opt:"local_addr"`
+	// Dialer affect connection when dialing an nsqd. Overwrite this to connect over proxy.
```
mreiferson, line 110: Let's use:

> Dialer will be called to connect to an nsqd when set, e.g. through a proxy

PR #305, opened 2020-08-06
command: register support topic/channel state

maplessssy:
command: register support topic/channel state
see [nsq issue:1199](https://github.com/nsqio/nsq/issues/1199)

### DIFF
--- a/command.go
+++ b/command.go
@@ -104,12 +104,12 @@ func Auth(secret string) (*Command, error) {
 }
 
 // Register creates a new Command to add a topic/channel with state for the connected nsqd
-func Register(topic string, channel string, paused ...int) *Command {
+func Register(topic string, channel string, states ...[]byte) *Command {
 	params := [][]byte{[]byte(topic), []byte(channel)}
-	if len(paused) > 0 {
-		params = append(params, []byte(strconv.Itoa(paused[0]))) //topic isPaused
-		if len(paused) > 1 {
-			params = append(params, []byte(strconv.Itoa(paused[1]))) //channel isPaused
+	if len(states) > 0 {
+		params = append(params, states[0]) //topic state
+		if len(states) > 1 {
+			params = append(params, states[1]) //channel state
 		}
 	}
 	return &Command{[]byte("REGISTER"), params, nil}
@@ -124,9 +124,9 @@ func UnRegister(topic string, channel string) *Command {
 	return &Command{[]byte("UNREGISTER"), params, nil}
 }
 
-// SyncState creates a new Command to sync a topic/channel pause state changes.
-func SyncState(topic string, channel string, state int) *Command {
-	params := [][]byte{[]byte(strconv.Itoa(state)), []byte(topic)}
+// SyncState creates a new Command to sync a topic/channel state changes, like paused, unpaused
+func SyncState(topic string, channel string, state []byte) *Command {
+	params := [][]byte{state, []byte(topic)}
 	if len(channel) > 0 {
 		params = append(params, []byte(channel))
 	}
--- a/command.go
+++ b/command.go
@@ -124,6 +124,7 @@ func UnRegister(topic string, channel string) *Command {
 	return &Command{[]byte("UNREGISTER"), params, nil}
 }
 
+// SyncState creates a new Command to sync a topic/channel pause state changes.
 func SyncState(topic string, channel string, state int) *Command {
 	params := [][]byte{[]byte(strconv.Itoa(state)), []byte(topic)}
 	if len(channel) > 0 {
--- a/command.go
+++ b/command.go
@@ -104,12 +104,13 @@ func Auth(secret string) (*Command, error) {
 }
 
 // Register creates a new Command to add a topic/channel with state for the connected nsqd
-func Register(topic string, topicPaused int, channel string, channelPaused int) *Command {
-	params := [][]byte{
-		[]byte(topic),
-		[]byte(channel),
-		[]byte(strconv.Itoa(topicPaused)),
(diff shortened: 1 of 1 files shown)


thekingofworld:
hi, anyone can help to take a look at this?

ploxiln:
Yeah, yeah, I saw it.

The interesting detail, is how current nsqlookupd will handle receiving more params. Backwards compatibility is needed while rolling out new versions. It looks like current nsqlookupd will ignore extra params, which is good.

There is another interesting detail: currently, topics/channels are only registered once, per connection to nsqlookupd, I think. But to use this, they will need to be re-registered whenever pause state changes. That's a possibly interesting change. There's also the fact that the topic pause state is sent redundantly when every channel state is sent ... not necessarily a problem for performance/efficiency, but rather confusion about what to do with multiple potentially conflicting reports. Always take the last topic state reported, even if that was just part of a channel state report/registration, I guess?

I think it will all work out fine, but it is good to clearly see how the details will look in the other projects, before finalizing this design. So, I would open up a pull request on the main nsq repo, modifying nsqd and nsqlookupd to use this new functionality, so we can see how that looks. You can try to temporarily modify the go.mod to use this commit of go-nsq so your nsq PR is able to build and test.

thekingofworld:
@ploxiln I hava opend up a pull request on the main nsq repo, see [nsq pr](https://github.com/nsqio/nsq/pull/1274)

mreiferson:
Building on @ploxiln's feedback, I think the underlying problem with this implementation is that it's not a "registration". Perhaps we need a new command that represents a topic/channel "state change"?

thekingofworld:
> Building on @ploxiln's feedback, I think the underlying problem with this implementation is that it's not a "registration". Perhaps we need a new command that represents a topic/channel "state change"?

I think a new command is good, it's more clearly that topic and channel state change can delivery more independently. @ploxiln and also resolve the topic pause state is sent redundantly when every channel state is sent .

command.go
```diff
@@ -121,6 +124,15 @@ func UnRegister(topic string, channel string) *Command {
 	return &Command{[]byte("UNREGISTER"), params, nil}
 }
 
+// SyncState creates a new Command to sync a topic/channel pause state changes.
+func SyncState(topic string, channel string, state int) *Command {
```
mreiferson, line 128: This command should probably send a JSON-serialized dictionary of key/values, e.g.

```json
{
    "paused": true
}
```

command.go
```diff
@@ -103,11 +103,14 @@ func Auth(secret string) (*Command, error) {
 	return &Command{[]byte("AUTH"), nil, []byte(secret)}, nil
 }
 
-// Register creates a new Command to add a topic/channel for the connected nsqd
-func Register(topic string, channel string) *Command {
-	params := [][]byte{[]byte(topic)}
-	if len(channel) > 0 {
-		params = append(params, []byte(channel))
+// Register creates a new Command to add a topic/channel with state for the connected nsqd
+func Register(topic string, channel string, paused ...int) *Command {
```
mreiferson, line 107: Should probably do the same here? (send a JSON-serialized dictionary of key/values as the third argument)

command.go
thekingofworld: Yeah, I think it's all about extendability, by Json, we can use multiple different key-value pairs to represent different states.
What I thought at the time was that we can use integer value to represent different states, like 0 for unpaused, 1 for paused, 2,3,4...for else.
Maybe Json extendability is better? and also readability.

thekingofworld: agree change to json.  : )

thekingofworld:
@mreiferson This PR is ready for review. PTAL.
/cc @ploxiln

command.go
```diff
@@ -103,11 +103,14 @@ func Auth(secret string) (*Command, error) {
 	return &Command{[]byte("AUTH"), nil, []byte(secret)}, nil
 }
 
-// Register creates a new Command to add a topic/channel for the connected nsqd
-func Register(topic string, channel string) *Command {
-	params := [][]byte{[]byte(topic)}
-	if len(channel) > 0 {
-		params = append(params, []byte(channel))
+// Register creates a new Command to add a topic/channel with state for the connected nsqd
+func Register(topic string, channel string, states ...[]byte) *Command {
```
mreiferson, line 107: I'm still not particularly happy with this API.

For one, it's a breaking change to `go-nsq` (yes, I realize that it's incredibly unlikely that users are depending on this function to speak to `nsqlookupd` over the TCP protocol, nonetheless).

We could avoid introducing this breaking change by not touching `Register` at all, instead just _adding_ the new `SyncState` command and then sending a `SyncState` command immediately following a new registration (in `nsqd`).

command.go
```diff
@@ -121,6 +124,15 @@ func UnRegister(topic string, channel string) *Command {
 	return &Command{[]byte("UNREGISTER"), params, nil}
 }
 
+// SyncState creates a new Command to sync a topic/channel state changes, like paused, unpaused
+func SyncState(topic string, channel string, state []byte) *Command {
```
mreiferson, line 128: I would expect this function to accept `state` as a `map[string]interface{}`, rather than `[]byte` and this function would handle serialization (like `Identify`'s implementation above).

command.go
thekingofworld: Okay, I removed what I added before

command.go
thekingofworld: done, good suggestion

mreiferson:
this is looking good, thanks for your patience!

mreiferson (approved):
LGTM, but let's not merge this until we're done with https://github.com/nsqio/nsq/pull/1274

PR #313, opened 2020-11-23
Use AuthSecret as Bearer token for lookupd queries

jehiah:
The [`AUTH` protocol spec](https://nsq.io/clients/tcp_protocol_spec.html#auth)  and [nsqd AUTH details](https://nsq.io/components/nsqd.html#auth) don't give much guidance for how to use AUTH with a lookupd endpoint (except implying that [`nsqauthfilter`](https://github.com/jehiah/nsqauth-contrib#nsqauthfilter) is one approach).

Currently the only exposed capability to use a lookup endpoint that requires authorization is to bake authorization credentials into the URL used as a lookup endpoint. go-nsq then makes a GET request to that endpoint during discovery.

Security best practice is to move authorization details to a header and out of the URL i.e. a `Authorization: Bearer {Auth Secret}` header.

This issue is to introduce new behavior where an AuthSecret when set is used as a Bearer token on lookupd calls. That seems like a sane default

`SkipLookupdAuthorization` can be used to opt into the current behavior (AuthSecret not used on lookupd requests)

cc: @mreiferson @ploxiln  in case you have any thoughts on this approach. I think our AUTH docs could use a little help, i'll try to build some better documentation of that.

### DIFF
--- a/config.go
+++ b/config.go
@@ -176,8 +176,10 @@ type Config struct {
 	// The server-side message timeout for messages delivered to this client
 	MsgTimeout time.Duration `opt:"msg_timeout" min:"0"`
 
-	// secret for nsqd authentication (requires nsqd 0.2.29+)
+	// Secret for nsqd authentication (requires nsqd 0.2.29+)
 	AuthSecret string `opt:"auth_secret"`
+	// Use AuthSecret as 'Authorization: Bearer {AuthSecret}' on lookupd queries
+	LookupdAuthorization bool `opt:"skip_lookupd_authorization" default:"true"`
 }
 
 // NewConfig returns a new default nsq configuration.
--- a/api_request.go
+++ b/api_request.go
@@ -3,7 +3,6 @@ package nsq
 import (
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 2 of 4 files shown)


ploxiln (approved):
Makes sense to me.

(I did take a moment to remind myself of the elements here: the AuthSecret is currently just used for tcp-protocol to nsqd, and with this change it still is, but with the addition of being sent to nsqlookupd in the auth header, unless explicitly disabled. Should be fully backward compatible, except it can "expose" your AuthSecret to your nsqlookupd by default, which is probably not the end of the world.)

config.go
```diff
@@ -178,6 +178,8 @@ type Config struct {
 
 	// secret for nsqd authentication (requires nsqd 0.2.29+)
 	AuthSecret string `opt:"auth_secret"`
+	// skip using AuthSecret as 'Authorization: Bearer {AuthSecret}' on lookupd queries
+	SkipLookupdAuthorization bool `opt:"skip_lookupd_authorization"`
```
mreiferson, line 182: Rather than having a double-negative to enable, why not call this `LookupdAuthorization` and default it to `true`?

mreiferson:
LGTM otherwise

config.go
jehiah: My intention was to structure a setting where the zero value was default; While this struct is always created through `NewConfig` it still seems like good practice to leverage the zero value.

`LookupdAuthorization` is also a good name though, so i don't have a strong objection here. thoughts?

PR #314, opened 2020-12-07
Check nsqlookupd list emptiness on Consumer instantiation

borodiychuk:
If empty nsqlookupd servers list is `ConnectToNSQLookupds` is provided, then Consumer instantiation does not fail, which is confusing. This PR makes it return an error for such case.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -374,6 +374,9 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
 //
 // A goroutine is spawned to handle continual polling.
 func (r *Consumer) ConnectToNSQLookupds(addresses []string) error {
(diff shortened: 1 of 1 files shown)


ploxiln:
It is a somewhat common pattern to sometimes call this with an empty list, as part of allowing _either_ nsqd-tcp-address _or_ lookupd-http-address lists:

https://github.com/nsqio/nsq/blob/master/apps/nsq_tail/nsq_tail.go#L124

https://github.com/nsqio/nsq/blob/master/apps/nsq_to_http/nsq_to_http.go#L289

borodiychuk:
Oh, I see, it makes sense, yes.

ploxiln:
It's reasonable to question :)

PR #317, opened 2021-01-15
producer: improve code readability

martin-sucha:
Removing empty switch case makes the code more readable.
See https://github.com/nsqio/go-nsq/pull/310#discussion_r517885032


mreiferson:
thanks

Issue #316, opened 2021-01-12
panic in Consumer.nextLookupdEndpoint


martin-sucha:
Hi,

we have seen a panic like this when an invalid address is provided to `ConnectToNSQLookupd` with `go-nsq@v1.0.8`:

```
panic: parse "http:// myhostname:4161": invalid character " " in host name
goroutine 1870 [running]:
github.com/nsqio/go-nsq.(*Consumer).nextLookupdEndpoint(0xc00034ab00, 0x1, 0x139bd85)
/go/pkg/mod/github.com/nsqio/go-nsq@v1.0.8/consumer.go:432 +0x477
github.com/nsqio/go-nsq.(*Consumer).queryLookupd(0xc00034ab00)
/go/pkg/mod/github.com/nsqio/go-nsq@v1.0.8/consumer.go:467 +0x96
github.com/nsqio/go-nsq.(*Consumer).lookupdLoop(0xc00034ab00)
/go/pkg/mod/github.com/nsqio/go-nsq@v1.0.8/consumer.go:397 +0x34e
created by github.com/nsqio/go-nsq.(*Consumer).ConnectToNSQLookupd
/go/pkg/mod/github.com/nsqio/go-nsq@v1.0.8/consumer.go:340 +0x1c7
```

I would expect `ConnectToNSQLookupd` to return error in this case instead of the code panicking later.

There is some validation in `validatedLookupAddr` but the code is different than the one in `nextLookupdEndpoint`.

I guess the best fix would be to parse the address just once in `ConnectToNSQLookupd` and store the parsed URL into `r.lookupdHTTPAddrs`

mreiferson:
Agreed, we should error earlier than the connect loop / goroutine. Would you be interested in submitting a fix for this?

PR #321, opened 2021-05-15
consumer: fix panic in nextLookupdEndpoint

martin-sucha:
The code can panic if invalid lookupd address is provided:

    panic: parse "http:// myhostname:4161": invalid character " " in host name

This is caused different validation of the input when configured
in ConnectToNSQLookupd and when used in nextLookupdEndpoint.

Moving preparation of the URL to the beginning so that we don't
need to do any input validation in nextLookupEndpoint.

Fixes https://github.com/nsqio/go-nsq/issues/316

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -340,7 +340,8 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
 		return errors.New("no handlers")
 	}
 
-	if err := validatedLookupAddr(addr); err != nil {
+	parsedAddr, err := buildLookupAddr(addr, r.topic)
+	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


martin-sucha:
Note that I have not tested the code yet as I don't have a nsq instance on the machine I'm currently on.

mreiferson:
I'm confused, it looks like `validatedLookupAddr` already returned an error from `url.Parse`, perhaps in your test case you're not checking the return value from `ConnectToNSQLookupd`?

martin-sucha:
If you use `" myhostname:4161"` as the input value then `validatedLookupAddr` does not call `url.Parse` at all because the string does not contain a slash.

martin-sucha:
I've updated the commit message so that it is clearer.

consumer.go
```diff
@@ -126,7 +126,7 @@ type Consumer struct {
 
 	// used at connection close to force a possible reconnect
 	lookupdRecheckChan chan int
-	lookupdHTTPAddrs   []string
+	lookupdHTTPAddrs   []lookupdHTTPAddr
```
mreiferson, line 129: I don't see any benefit of introducing a type here, let's keep this a `string` and then a few of the other changes here won't be necessary.

mreiferson (changes requested):
Thanks @martin-sucha, I think we can simplify this a bit so the diff isn't as noisy.

consumer.go
martin-sucha: Ok, I removed the type.

PR #326, opened 2021-06-25
fix send closed channel when consumer reconnecting

Allenxuxu:


mreiferson:
Thanks for submitting this @Allenxuxu!

The `Consumer` exit code is unfortunately extremely convoluted and delicate, and I'm not convinced this fully addresses the issue. I think what needs to happen (at a high-level) is:

1. `Stop()`
2. Prevent reconnections (for both "lookup mode" and "direct connection") by ensuring that we exit `lookupdLoop` and complete any reconnect goroutines (like you've done here)
3. Then we need to wait on all connections to close
4. Stop handlers

PR #333, opened 2021-09-28
feat: reuse http client

JieTrancender:
### Problem
After some wrong happend on my machine, i found there are many `TIME_WAIT` status. 
I have many nsqd topics and n consumers, consumer making http request which not using `Keep-Alive` causes this problem.
~~~
[xxx@xxx-inc.com server]$ netstat -an |grep 19991|awk '{print $6}'|sort|uniq -c
      2 ESTABLISHED
      1 LISTEN
   1401 TIME_WAIT
~~~

### Solution
1. using go http.Transport features(keepalives,timeouts,dualstack)
2. reuse http client
~~~
[xxx@xxx-inc.com server]$ netstat -an |grep 19991|awk '{print $6}'|sort|uniq -c
   1217 ESTABLISHED
      1 LISTEN
    305 TIME_WAIT
~~~

3. support setting lookupd http client
~~~
[xxx@xxx-inc.com server]$ netstat -an |grep 19991|awk '{print $6}'|sort|uniq -c
    131 ESTABLISHED
      1 LISTEN
    355 TIME_WAIT
~~~

### DIFF
--- a/api_request.go
+++ b/api_request.go
@@ -26,15 +26,17 @@ func (c *deadlinedConn) Write(b []byte) (n int, err error) {
 
 func newDeadlineTransport(timeout time.Duration) *http.Transport {
 	transport := &http.Transport{
-		DisableKeepAlives: true,
-		Dial: func(netw, addr string) (net.Conn, error) {
-			c, err := net.DialTimeout(netw, addr, timeout)
-			if err != nil {
-				return nil, err
-			}
-			return &deadlinedConn{timeout, c}, nil
-		},
+		DialContext: (&net.Dialer{
+			Timeout:   timeout,
+			KeepAlive: 60 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		ResponseHeaderTimeout: timeout,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
 	}
+
 	return transport
 }
 
--- a/api_request.go
+++ b/api_request.go
@@ -47,8 +47,7 @@ type wrappedResp struct {
 }
 
 // stores the result in the value pointed to by ret(must be a pointer)
-func apiRequestNegotiateV1(method string, endpoint string, headers http.Header, ret interface{}) error {
-	httpclient := &http.Client{Transport: newDeadlineTransport(2 * time.Second)}
+func apiRequestNegotiateV1(httpclient *http.Client, method string, endpoint string, headers http.Header, ret interface{}) error {
 	req, err := http.NewRequest(method, endpoint, nil)
 	if err != nil {
 		return err
--- a/config.go
+++ b/config.go
@@ -108,8 +108,10 @@ type Config struct {
 	//
 	// NOTE: when not using nsqlookupd, LookupdPollInterval represents the duration of time between
 	// reconnection attempts
-	LookupdPollInterval time.Duration `opt:"lookupd_poll_interval" min:"10ms" max:"5m" default:"60s"`
-	LookupdPollJitter   float64       `opt:"lookupd_poll_jitter" min:"0" max:"1" default:"0.3"`
+	LookupdPollInterval      time.Duration `opt:"lookupd_poll_interval" min:"10ms" max:"5m" default:"60s"`
+	LookupdPollJitter        float64       `opt:"lookupd_poll_jitter" min:"0" max:"1" default:"0.3"`
+	LookupdPollAliveDuration time.Duration `opt:"lookupd_pool_alive_duration" default:"60s"`
+	LookupdPollTimeout       time.Duration `opt:"lookupd_poll_timeout" default:"6s"`
 
 	// Maximum duration when REQueueing (for doubling of deferred requeue)
 	MaxRequeueDelay     time.Duration `opt:"max_requeue_delay" min:"0" max:"60m" default:"15m"`
--- a/consumer.go
+++ b/consumer.go
@@ -327,6 +327,11 @@ func (r *Consumer) ChangeMaxInFlight(maxInFlight int) {
 	}
 }
 
+// set lookupd http client
+func (r *Consumer) SetLookupdHttpClient(httpclient *http.Client) {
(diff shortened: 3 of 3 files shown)


api_request.go
```diff
@@ -26,15 +26,17 @@ func (c *deadlinedConn) Write(b []byte) (n int, err error) {
...
-		},
+		DialContext: (&net.Dialer{
+			Timeout:   timeout,
+			KeepAlive: 60 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		ResponseHeaderTimeout: timeout,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
```
ploxiln, line 37: I think this whole function `newDeadlineTransport()` can be deleted, as `apiRequestNegotiateV1()` was the only user, and now it uses the passed `http.Client` instead

config.go
```diff
@@ -108,8 +108,10 @@ type Config struct {
 	//
 	// NOTE: when not using nsqlookupd, LookupdPollInterval represents the duration of time between
 	// reconnection attempts
-	LookupdPollInterval time.Duration `opt:"lookupd_poll_interval" min:"10ms" max:"5m" default:"60s"`
-	LookupdPollJitter   float64       `opt:"lookupd_poll_jitter" min:"0" max:"1" default:"0.3"`
+	LookupdPollInterval      time.Duration `opt:"lookupd_poll_interval" min:"10ms" max:"5m" default:"60s"`
+	LookupdPollJitter        float64       `opt:"lookupd_poll_jitter" min:"0" max:"1" default:"0.3"`
+	LookupdPollAliveDuration time.Duration `opt:"lookupd_pool_alive_duration" default:"60s"`
```
ploxiln, line 113: naming idea ... Duration may be a bit redundant, and KeepAlive more clear

```suggestion
	LookupdPollKeepalive time.Duration `opt:"lookupd_pool_keepalive" default:"60s"`
```

consumer.go
```diff
@@ -355,6 +361,24 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
 		}
 	}
 	r.lookupdHTTPAddrs = append(r.lookupdHTTPAddrs, parsedAddr)
+	if r.lookupdHttpClient == nil {
+		transport := &http.Transport{
+			DialContext: (&net.Dialer{
+				Timeout:   r.config.LookupdPollTimeout,
+				KeepAlive: r.config.LookupdPollAliveDuration,
```
ploxiln, line 368: This is the TCP keepalive feature, not the http keepalive feature (which is enabled by default, just need to reuse the transport, instead of making a new one every time). This may not be needed:

https://pkg.go.dev/net#Dialer
```
	// KeepAlive specifies the interval between keep-alive
	// probes for an active network connection.
	// If zero, keep-alive probes are sent with a default value
	// (currently 15 seconds), if supported by the protocol and operating
	// system. Network protocols or operating systems that do
	// not support keep-alives ignore this field.
	// If negative, keep-alive probes are disabled.
	KeepAlive time.Duration
```

consumer.go
```diff
@@ -355,6 +361,24 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
...
+				DualStack: true,
```
ploxiln, line 369: looks like this also isn't needed:

```
	// DualStack previously enabled RFC 6555 Fast Fallback
	// support, also known as "Happy Eyeballs", in which IPv4 is
	// tried soon if IPv6 appears to be misconfigured and
	// hanging.
	//
	// Deprecated: Fast Fallback is enabled by default. To
	// disable, set FallbackDelay to a negative value.
	DualStack bool
```

consumer.go
ploxiln: hmm ... the `DefaultTransport` does still seem to explicitly configure this to 30 seconds ... we should probably similarly set it to 30 seconds, but not bother with making it configurable (it's the tcp layer thing, not http or nsq logic related, 30 seconds is probably fine everywhere, and the original problem was creating a new Transport for every request)

config.go
```diff
@@ -108,8 +108,10 @@ type Config struct {
...
+	LookupdPollTimeout       time.Duration `opt:"lookupd_poll_timeout" default:"6s"`
```
ploxiln, line 114: we might need to make this longer by default ... apparently some clusters have thousands of topics and channels which can slow down nsqlookupd pretty badly: https://github.com/nsqio/nsq/pull/1208

consumer.go
JieTrancender: Yes, it's not necessary. At first i just wanted him to clear a little bit, and i have removed it.

JieTrancender:
@ploxiln Thank you for your review, i have finished those modification, please review it again.

UPGRADING.md
```diff
@@ -173,7 +173,7 @@ conforming to standard library logging conventions.
 
 #### Misc.
 
-Un-exported `NewDeadlineTransport` and `ApiRequest`, which never should have been exported in the
+Un-exported `ApiRequest`, which never should have been exported in the
```
ploxiln, line 176: you can leave this as it was; it's still true that `NewDeadlineTransport` was un-exported (and then now has been un-existed ;)

consumer.go
```diff
@@ -355,6 +361,23 @@ func (r *Consumer) ConnectToNSQLookupd(addr string) error {
 		}
 	}
 	r.lookupdHTTPAddrs = append(r.lookupdHTTPAddrs, parsedAddr)
+	if r.lookupdHttpClient == nil {
+		transport := &http.Transport{
+			DialContext: (&net.Dialer{
+				Timeout:   r.config.LookupdPollTimeout,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			ResponseHeaderTimeout: r.config.LookupdPollTimeout,
```
ploxiln, line 370: This seems redundant with the `http.Client` `Timeout` but I guess it's fine.
(Dialer Timeout above could probably be a shorter default, probably matching the 10s TLSHandshakeTimeout below. But again, I guess it's fine, doesn't really hurt.)

PR #337, opened 2021-10-19
rename module to github.com/CodapeWild/go-nsq

CodapeWild:


ploxiln:
It looks like you intended to open this PR against your own repo, rather than this upstream one.

PR #346, opened 2022-06-05
producer: handle case of no transactions in popTransaction()

### LINKED ISSUE #345: Panic in producer's router
We have retries system which does resending to producers in a round-robin manner, and under some circumstances when nsq services are restarting - this library can easily crash the application. It looks like its internals are not thread-safe and written in a very outdated golang style, which makes it extremely hard to investigate where exactly issue is:

go-nsq: v1.1.0 

```
panic: runtime error: index out of range [0] with length 0

goroutine 48061 [running]:
github.com/nsqio/go-nsq.(*Producer).popTransaction(...)
	/go/src/***/vendor/github.com/nsqio/go-nsq/producer.go:370
github.com/nsqio/go-nsq.(*Producer).router(0xc0003c18c0)
	/go/src/***/vendor/github.com/nsqio/go-nsq/producer.go:353 +0x532
created by github.com/nsqio/go-nsq.(*Producer).connect
	/go/src/***/vendor/github.com/nsqio/go-nsq/producer.go:324 +0x574
```

PS. please dont fork goroutines uncontrollably, instead of poorly controlled channels - use a friendly worker pool library with a managed set of worker goroutines, where you can wait for all tasks to be done, and stop using goto - its not ANSI C89...

ploxiln:
This corner case generally should not happen, but a nsqd bug or
strange network condition could possibly send erroneous bytes that
are interpreted as a response. The Producer should be robust,
and not panic the whole process.

fixes #345 (hopefully)

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -367,6 +367,17 @@ exit:
 }
 
 func (w *Producer) popTransaction(frameType int32, data []byte) {
+	if len(w.transactions) == 0 {
+		dataLen := len(data)
+		if dataLen > 32 {
+			data = data[:32]
+		}
+		w.log(LogLevelError,
+			"(%s) unexpected response type=%d len=%d data[:32]=0x%x",
+			w.conn.String(), frameType, dataLen, data)
+		w.close()
+		return
+	}
 	t := w.transactions[0]
 	w.transactions = w.transactions[1:]
 	if frameType == FrameTypeError {


mreiferson:
yea, it makes sense to be defensive here, but let's also log the `data` so if/when we catch this in the wild we better understand what's happening?

yiippee:
@ploxiln  You said the producer is "thread-safe", but how it is guaranteed to get responses in the order which commands were sent?  The nsqd guarantees it? Thanks.
```
func (w *Producer) popTransaction(frameType int32, data []byte) {
	t := w.transactions[0]  // How can you be sure it's the first one element.  
	w.transactions = w.transactions[1:]
	if frameType == FrameTypeError {
		t.Error = ErrProtocol{string(data)}
	}
	t.finish()
}
```

ploxiln:
> how it is guaranteed to get responses in the order which commands were sent? The nsqd guarantees it?

Yes, in this way it's similar to http request "pipelining"

mreiferson:
@ploxiln can you update this as per https://github.com/nsqio/go-nsq/pull/346#issuecomment-1152959700

ploxiln:
OK - how does that look? (first 32 bytes in hex)

mreiferson (approved):

PR #350, opened 2022-11-10
fix(consumer): remove old nsqd connections if addresses change

jrkt:
Right now, in the event that nsqd addresses change, the old connections stay around in the consumer's connections map. This simply cleans up old connections from that map.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -33,9 +33,9 @@ type Handler interface {
 // HandlerFunc is a convenience type to avoid having to declare a struct
 // to implement the Handler interface, it can be used like this:
 //
-// 	consumer.AddHandler(nsq.HandlerFunc(func(m *Message) error {
-// 		// handle the message
-// 	}))
+//	consumer.AddHandler(nsq.HandlerFunc(func(m *Message) error {
+//		// handle the message
+//	}))
 type HandlerFunc func(message *Message) error
 
 // HandleMessage implements the Handler interface
@@ -220,8 +220,7 @@ func (r *Consumer) conns() []*Conn {
 // The logger parameter is an interface that requires the following
 // method to be implemented (such as the the stdlib log.Logger):
 //
-//    Output(calldepth int, s string) error
-//
+//	Output(calldepth int, s string) error
 func (r *Consumer) SetLogger(l logger, lvl LogLevel) {
 	r.logGuard.Lock()
 	defer r.logGuard.Unlock()
@@ -266,8 +265,7 @@ func (r *Consumer) getLogLevel() LogLevel {
 // of the following interfaces that modify the behavior
 // of the `Consumer`:
 //
-//    DiscoveryFilter
-//
+//	DiscoveryFilter
 func (r *Consumer) SetBehaviorDelegate(cb interface{}) {
 	matched := false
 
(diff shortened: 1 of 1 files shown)


jrkt:
@mreiferson could I get some thoughts on this?

consumer.go
```diff
@@ -33,9 +33,9 @@ type Handler interface {
 // HandlerFunc is a convenience type to avoid having to declare a struct
 // to implement the Handler interface, it can be used like this:
 //
-// 	consumer.AddHandler(nsq.HandlerFunc(func(m *Message) error {
-// 		// handle the message
-// 	}))
+//	consumer.AddHandler(nsq.HandlerFunc(func(m *Message) error {
+//		// handle the message
+//	}))
```
ploxiln, line 38: it looks like these comment change are from go-1.19 gofmt changes - can you separate them into a different pull request?

consumer.go
```diff
@@ -513,13 +511,33 @@ retry:
...
 		if err != nil && err != ErrAlreadyConnected {
 			r.log(LogLevelError, "(%s) error connecting to nsqd - %s", addr, err)
 			continue
 		}
+		successfulNsqdAddrs = append(successfulNsqdAddrs, addr)
 	}
+
+	// in the event that there are new nsqd addresses, remove the old connections from the connections map
+	for addr := range r.connections {
+		if !inAddrs(successfulNsqdAddrs, addr) {
```
ploxiln, line 527: These successfulNsqdAddrs from a single lookupd. It is expected that the consumer effectively connects to the union of the results from multiple lookupd. Often, all lookupd return the same results, but it is expected that while changes to the set of nsqlookupd propagate there will be inconsistencies, and that's OK because of the "union" behavior. The behavior is also supposed to be resilient to some temporary network disruptions. So we don't really want to remove/disconnect nsqds which are missing from a single lookupd response.

ploxiln:
Normally, connections are removed from `Consumer.connections` in `Consumer.onConnClose()`, which does various other cleanup tasks, to keep things consistent.

Are you having an issue where the tcp connection just hangs and is never detected to be closed?

consumer.go
jrkt: But the polling lookup loop only happens on the first lookupd address. https://github.com/nsqio/go-nsq/blob/b8bb2c58cd0829d58d66718d8d4dd6c925ca68b6/consumer.go#L383. We are having issues when the upstream IPs change and simply looking at the code it seemed troublesome that if the IPs change, old connections are not closed.

jrkt:
@ploxiln when does `Consumer.onConnClose` get called? It looks like it only gets called when you close the consumer but what happens to connections that don't point to a valid IP in the event that upstream IPs change?

ploxiln:
When the connection has some error, including detecting TCP close, it goes through a sequence:
https://github.com/nsqio/go-nsq/blob/master/conn.go#L640
that eventually calls `consumerConnDelegate.OnClose()`:
https://github.com/nsqio/go-nsq/blob/master/conn.go#L725
That calls `Consumer.onConnClose()`:
https://github.com/nsqio/go-nsq/blob/0e8d7a73ef5e81b6c3e2bdc4a9a5fb9afa44a36b/delegates.go#L121

In cases where a server really just switches ip addresses abruptly, there is no TCP FIN or RST that can be sent from the original IP to close the connection, so it'll have to time-out eventually, but operating systems may not enable TCP "keep-alives" by default. But if you do enable them, they should cause the zombie connection to be closed in a few minutes.
https://github.com/golang/go/commit/5bd7e9c54f946eec95d32762e7e9e1222504bfc1
https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html

consumer.go
ploxiln: When the first lookupd address is added, this `queryLookupd()` is called directly, then the `lookupdLoop()` is spawned to the background. That loop calls this `queryLookupd()` again periodically, while there usually are multiple lookupd addresses, and one of them is queried at a time by this function in a round-robin fashion.

PR #353, opened 2023-01-12
Limit read message size

### LINKED ISSUE #352: producer: HTTP response to IDENTIFY causes memory allocation of over 1GB
### Problem description

If we create a producer pointing to the HTTP address of nsqd by mistake (instead of using the TCP endpoint), the producer allocates over 1GB of memory before failing. This happens when the nsq protocol response size is extracted from the first 4 bytes of the tcp response data:

https://github.com/nsqio/go-nsq/blob/259dc590b2e791e4b11ad7496bc312ec81d1004b/protocol.go#L55

The first 4 bytes of `HTTP/1.1 400 Bad Request` produce the int `1213486160`, which is used to create the buffer to hold the nsq response data:

https://github.com/nsqio/go-nsq/blob/259dc590b2e791e4b11ad7496bc312ec81d1004b/protocol.go#L64

And boom, another OOMKilled pod.

### How to reproduce

This code can be used to reproduce the issue. You'll need an instance nsqd running locally.

```go
package main

import (
	"log"
	"runtime"

	"github.com/nsqio/go-nsq"
)

func main() {
	config := nsq.NewConfig()
	p, err := nsq.NewProducer("localhost:4151", config)
	if err != nil {
		log.Fatalln("NewProducer:", err)
	}

	err = p.Publish("test-topic", []byte("test-message"))
	printTotalAlloc()
	if err != nil {
		log.Fatalln("Publish:", err)
	}
	log.Println("Finished")
}

func printTotalAlloc() {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	log.Printf("TotalAlloc: %.3f MiB", float64(m.TotalAlloc)/1000/1000)
}
```

### Expected behaviour

I guess the producer should somehow make sure that the nsq protocol is being used before trying to deserialize the data, but I'm not familiar with the nsq protocol.

### Evidences

Running the code above prints:

```
2023/01/12 13:47:51 INF    1 (localhost:4151) connecting to nsqd
2023/01/12 13:47:51 ERR    1 (localhost:4151) error connecting to nsqd - failed to IDENTIFY - unexpected EOF
2023/01/12 13:47:51 TotalAlloc: 1213.861 MiB
2023/01/12 13:47:51 Publish: failed to IDENTIFY - unexpected EOF
```

### Context

I'm using go-nsq v1.1.0 and Go 1.19.

ibice:
- Add MaxMsgSize to configuration mimicking the nsqd cofiguration key. It defaults to 1048576 as the nsqd config -- source: https://github.com/nsqio/nsq/blob/a4939964f6715edd27a6904b87c2f9eb6a45e749/nsqd/options.go#L130

- Pass MaxMsgSize to ReadResponse via its caller, ReadUnpackedResponse. Check msgSize does not exceed the maximum in ReadResponse. This reduces the risk of attempting to read arbitrary (non-nsq) responses.

- Generate custom error if msgSize is the result of deserializing the first 4 bytes of an HTTP response (1213486160) to facilitate troublehooting

Fixes #352

### DIFF
--- a/config.go
+++ b/config.go
@@ -177,6 +177,9 @@ type Config struct {
 	// The server-side message timeout for messages delivered to this client
 	MsgTimeout time.Duration `opt:"msg_timeout" min:"0"`
 
+	// Maximum size of a single message in bytes (0 means no limit)
+	MaxMsgSize int32 `opt:"max_msg_size" min:"0" default:"0"`
+
 	// Secret for nsqd authentication (requires nsqd 0.2.29+)
 	AuthSecret string `opt:"auth_secret"`
 	// Use AuthSecret as 'Authorization: Bearer {AuthSecret}' on lookupd queries
--- a/protocol.go
+++ b/protocol.go
@@ -21,6 +21,9 @@ const (
 	FrameTypeMessage  int32 = 2
 )
 
+// Used to detect if an unexpected HTTP response is read
+const httpResponseMsgSize = 1213486160
+
 var validTopicChannelNameRegex = regexp.MustCompile(`^[\.a-zA-Z0-9_-]+(#ephemeral)?$`)
 
 // IsValidTopicName checks a topic name for correctness
@@ -48,7 +51,7 @@ func isValidName(name string) bool {
 //    |  4-byte  || N-byte
 //    ------------------------...
 //        size       data
-func ReadResponse(r io.Reader) ([]byte, error) {
+func ReadResponse(r io.Reader, maxMsgSize int32) ([]byte, error) {
 	var msgSize int32
 
 	// message size
@@ -60,6 +63,14 @@ func ReadResponse(r io.Reader) ([]byte, error) {
 	if msgSize < 0 {
 		return nil, fmt.Errorf("response msg size is negative: %v", msgSize)
 	}
+
+	if maxMsgSize > 0 && msgSize > maxMsgSize {
+		if msgSize == httpResponseMsgSize {
+			return nil, fmt.Errorf("unexpected HTTP response, a nsqd TCP endpoint is required")
+		}
+		return nil, fmt.Errorf("response msg size %v exceeds configured maximum (%v)", msgSize, maxMsgSize)
+	}
+
 	// message binary data
 	buf := make([]byte, msgSize)
 	_, err = io.ReadFull(r, buf)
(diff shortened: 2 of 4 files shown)


config.go
```diff
@@ -177,6 +177,9 @@ type Config struct {
 	// The server-side message timeout for messages delivered to this client
 	MsgTimeout time.Duration `opt:"msg_timeout" min:"0"`
 
+	// Maximum size of a single message in bytes
+	MaxMsgSize int32 `opt:"max_msg_size" min:"0" default:"1048576"`
```
mreiferson, line 181: I think it makes sense to have this config option, however introducing a new default size of 1M is going to be surprising for anyone upgrading `go-nsq` or using the built-ins we ship with the NSQ distribution like `nsqtofile`, etc.

I think this needs to default to `0`.

protocol.go
```diff
@@ -59,7 +59,12 @@ func ReadResponse(r io.Reader) ([]byte, error) {
 
 	if msgSize < 0 {
 		return nil, fmt.Errorf("response msg size is negative: %v", msgSize)
+	} else if msgSize == 1213486160 { // Integer value of deserializing "HTTP"
+		return nil, fmt.Errorf("unexpected HTTP response, a nsqd TCP endpoint is required")
+	} else if msgSize > maxMsgSize {
```
mreiferson, line 64: need to add a check here that `maxMsgSize > 0` too

mreiferson (changes requested):
Looks good so far, thanks for submitting the PR! Just a few comments...

config.go
ibice: I agree with your concern! I've set the default value of `MaxMsgSize` to 0.

protocol.go
ibice: Done! The new code looks like this:

```go
if msgSize < 0 {
	return nil, fmt.Errorf("response msg size is negative: %v", msgSize)
}

if maxMsgSize > 0 && msgSize > maxMsgSize {
	if msgSize == httpResponseMsgSize {
		return nil, fmt.Errorf("unexpected HTTP response, a nsqd TCP endpoint is required")
	}
	return nil, fmt.Errorf("response msg size %v exceeds configured maximum (%v)", msgSize, maxMsgSize)
}
```

The HTTP response detection logic won't get in the way, i.e: if by chance a message exactly as big as `1213486160` bytes is read and the configured limit is above that value (or `0`), the function won't fail.

mreiferson (approved):
LGTM, thanks! Would you mind squashing this down to 1 commit?

ibice:
> LGTM, thanks! Would you mind squashing this down to 1 commit?

Done @mreiferson!

ploxiln:
hmm flaky test, not sure if it's really this PR's fault ...

```
=== RUN   TestProducerHTTPConnectionFails
    producer_test.go:275: should detect unexpected HTTP response, but got err: failed to IDENTIFY - write tcp 127.0.0.1:45031->127.0.0.1:4151: write: broken pipe
--- FAIL: TestProducerHTTPConnectionFails (0.00s)
```

mreiferson:
thanks!

PR #365, opened 2024-05-14
producer: Integrate context.Context with publishing

### LINKED ISSUE #360: *: context.Context and timeouts
Go 1.7 added `context.Context` and it's been widely adopted for providing a way to control timeouts. 

This ticket is to track adding `context.Context` to `go-nsq` in a way that provides control over connections so the caller can avoid unexpected long operations (like a TCP dial timeout when trying to Publish)

This replaces #210 and #211 and along with #311 provide more reliable publishing in some edge cases.

At Bitly we have an interest in pursuing this change - we will follow up with a specific proposal with a mind to implementing in a backwards compatible way, and with a look towards easy transition to a future v2 that might drop calls w/o a context. We also want to be identify any changes needed in the design of #311 ahead of time before that change lands.

cc: @mreiferson @ploxiln @zoemccormick @Ulminator

Ulminator:
This PR sets out to resolve [#360](https://github.com/nsqio/go-nsq/issues/360) with the integration of [context.Context](https://pkg.go.dev/context#Context) in the `nsq.Producer`. In order to maintain backwards compatibility, new methods were added that have names with the suffix `WithContext`. These new methods mostly correspond to those that a client would use to publish messages and they adhere to context timeouts/deadlines. Once the deadline has been reached, an error will be returned to the client and the given `ProducerTransaction` will be dropped/not published, all while keeping the connection to `nsqd` alive.

An overview of the new methods added are below:

### nsq.Conn
* ConnectWithContext
* WriteCommandWithContext

### nsq.Producer
* PingWithContext
* PublishWithContext
* MultiPublishWithContext
* DeferredPublishWithContext
* PublishAsyncWithContext
* MultiPublishAsyncWithContext
* DeferredPublishAsyncWithContext

The idea here being that these `WithContext` methods would essentially replace the existing ones eventually, which would be a breaking/major version change for `go-nsq`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -184,7 +184,6 @@ func (c *Conn) ConnectWithContext(ctx context.Context) (*IdentifyResponse, error
 	// the timeout used is smallest of dialer.Timeout (config.DialTimeout) or context timeout
 	conn, err := dialer.DialContext(ctx, "tcp", c.addr)
 	if err != nil {
-		fmt.Println("dialer.DialContext error: ", err) // TODO: remove
 		return nil, err
 	}
 	c.conn = conn.(*net.TCPConn)
@@ -303,13 +302,11 @@ func (c *Conn) WriteCommand(cmd *Command) error {
 }
 
 func (c *Conn) WriteCommandWithContext(ctx context.Context, cmd *Command) error {
-	// would we want all of our usage of WriteCommand to be replaced with WriteCommandWithContext?
 	c.mtx.Lock()
 
 	var err error
 	select {
 	case <-ctx.Done():
-		fmt.Println("WriteCommandWithContext ctx.Done(): ", ctx.Err()) // TODO: remove
 		c.mtx.Unlock()
 		return ctx.Err()
 	default:
--- a/conn.go
+++ b/conn.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"compress/flate"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
@@ -119,8 +120,7 @@ func NewConn(addr string, config *Config, delegate ConnDelegate) *Conn {
 // The logger parameter is an interface that requires the following
 // method to be implemented (such as the the stdlib log.Logger):
 //
-//    Output(calldepth int, s string)
-//
+//	Output(calldepth int, s string)
 func (c *Conn) SetLogger(l logger, lvl LogLevel, format string) {
 	c.logGuard.Lock()
 	defer c.logGuard.Unlock()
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -291,13 +297,26 @@ func (c *Conn) Write(p []byte) (int, error) {
...
 
-	_, err := cmd.WriteTo(c)
-	if err != nil {
+	var err error
+	select {
+	case <-ctx.Done():
+		c.mtx.Unlock()
+		return ctx.Err()
+	default:
+		_, err := cmd.WriteTo(c)
```
mreiferson, line 313: I think we're shadowing `err` here and it won't be correct on line 323

producer.go
```diff
@@ -344,9 +388,17 @@ func (w *Producer) router() {
...
 			w.transactions = append(w.transactions, t)
-			err := w.conn.WriteCommand(t.cmd)
+			err := w.conn.WriteCommandWithContext(t.ctx, t.cmd)
 			if err != nil {
 				w.log(LogLevelError, "(%s) sending command - %s", w.conn.String(), err)
+				if err == context.Canceled || err == context.DeadlineExceeded {
+					// keep the connection alive if related to context timeout
+					// need to do some stuff that's in Producer.popTransaction here
+					w.transactions = w.transactions[1:]
+					t.Error = err
```
mreiferson, line 398: I don't love this code we've copied in to this context-specific edge case, can't we consolidate this in to `popTransaction`? Do we need to define a context-specific error code for the protocol?

mreiferson (changes requested):
Thanks for this, a few minor comments. I don't have any better ideas on naming functions in backwards-compatible way.

conn.go
Ulminator: Good catch! This should be fixed now 👍

producer.go
Ulminator: Yeah I wasn't a huge fan of how that was being handled...

I updated the code to invoke `popTransaction` without modifying the function signature, meaning I seemingly need to have at least one new "frameType" in `protocol.go` to handle the context errors. I opted to be explicit and add two "frameTypes" (for `context.Cancelled` and `context.DeadlineExceeded` respectively) so that we would not need to infer the context error from the byte slice/handle a default case like so:

```go
if frameType == FrameTypeContextError {
    switch string(data) {
    case context.Canceled.Error():
        t.Error = context.Canceled
    case context.DeadlineExceeded.Error():
        t.Error = context.DeadlineExceeded
    default:
        t.Error = ???
    }
}
```

The `FrameTypeContextCanceled` and `FrameTypeContextDeadlineExceeded` are not really "frameTypes" per se, so I also don't fully love this approach, but it feels better than what it was.

I'm curious to hear your thoughts on this approach and any suggestions you may have on how to improve upon this.

Ulminator:
@mreiferson thanks for the initial review!

Apologies it took so long for me to circle back around to this, but it should be good for another look 👍

martin-sucha:
Seems exactly like what I need, thanks!

I've reviewed the pull request and it looks good to me!

martin-sucha (approved):

producer.go
```diff
@@ -105,23 +109,27 @@ func NewProducer(addr string, config *Config) (*Producer, error) {
 // configured correctly, rather than relying on the lazy "connect on Publish"
 // behavior of a Producer.
 func (w *Producer) Ping() error {
+	ctx := context.Background()
+	return w.PingWithContext(ctx)
+}
+
+func (w *Producer) PingWithContext(ctx context.Context) error {
```
jehiah, line 123: We need to duplicate all of the docstrings onto the new methods

producer.go
jehiah: Can `popTransaction` take a 3rd `error` argument so we don't use magic values for some errors?

Ulminator:
Hey @jehiah ,

> Can popTransaction take a 3rd error argument so we don't use magic values for some errors?

I updated popTransaction to support this so now when there is a context error it is invoked like so:

```go
w.popTransaction(-1, []byte{}, err)
```

I set the first arg (`frameType`) to an invalid one since the only options are the following:

```go
const (
	FrameTypeResponse int32 = 0
	FrameTypeError    int32 = 1
	FrameTypeMessage  int32 = 2
)
```

Please let me know if I should change anything

